PR #37, opened 2016-09-29
[rules/eclipse*] slight improvement for eclipse packages/plugins

Cka3o4Huk:
Small fix for eclipse packages. 
Tested, at least works for me


AMDmi3:
Merged manually, thanks!

PR #38, opened 2016-09-29
Update README.md

Cka3o4Huk:
As of today repology.py requires py-requests (at least on FreeBSD).


AMDmi3:
I've updated README after latest changes, and this bit is already included. Maybe I need to push more frequently.

PR #42, opened 2016-09-30
Coalesce package names for adb/fastboot and gtest/gmock

jbeich:


rules.yaml
```diff
@@ -97,9 +99,12 @@
 - { name: xpi-adblock_plus,            setname: adblock-plus,             repos: [ freebsd ] }
 - { name: xpi-foxyproxy,               setname: foxyproxy,                repos: [ freebsd ] }
 - { name: zookeeper-c,                 setname: zookeeper,                repos: [ gentoo ]  }
+- { namepat: "android-platform-system-(core|extras)", setname: android-tools, repos: [ debian ] }
```
AMDmi3: Is this really valid while there's already android-tools package in Debian?

rules.yaml
```diff
@@ -97,9 +99,12 @@
...
+- { namepat: "android-tools-[^-]+",    setname: android-tools,            repos: [ freebsd, pkgsrc ] }
```
AMDmi3: This only applies to freebsd, no?

rules.yaml
```diff
@@ -97,9 +99,12 @@
...
 - { namepat: "celestia-.*",            setname: celestia,                 repos: [ freebsd ] } # merges glut/gnome/gtk, just GUI variants
 - { namepat: "eclipse-plugin-(.*)",    setname: "eclipse-$1",             repos: [ openbsd ] }
 - { namepat: "eiskaltdcpp-.*",         setname: eiskaltdcpp,              repos: [ freebsd ] }
+- { namepat: "g(oogle)?-?mock",        setname: gmock,                    repos: [ freebsd, debian ] }
```
AMDmi3: Better to just split it to two simple ```name:``` rules for freebsd and debian.

rules.yaml
jbeich: pkgsrc-wip also has [android-tools-adb](http://pkgsrc.se/wip/android-tools-adb).

rules.yaml
jbeich: adb is part of [platform/system/core](https://android.googlesource.com/platform/system/core/). Debian changed how they package other stuff from that repo, so android-tools is [deprecated](https://wiki.debian.org/AndroidTools/#Original_android-tools_package).

rules.yaml
jbeich: OK. Rebased.

rules.yaml
jbeich: Mapping Android stuff is going to be a mess either way:
- fastboot (like adb) lives in `core` repo but has dependencies in `extras`, `libselinux`.
- simpleperf lives in `extras` but has dependencies in `core`, `bionic` and `lzma` (in trunk).
- Fedora (and RPM-friends), Slackware, Gentoo, Arch have adb/fastboot in android-tools
- Arch and Debian also have adb/fastboot in *-platform-tools with Android SDK versioning scheme

PR #60, opened 2016-10-17
extend maps by ubuntu for lua- and python- prefixes

olevole:


AMDmi3:
debian repository type covers both debian and ubuntu, so this is not needed.

PR #105, opened 2017-01-02
initial pass at vagrant stuff

swills:


AMDmi3:
What is this?

swills:
Lets you use vagrant to bring up a system for running repology. Ansible will install all the packages needed for you, setup a git working copy of the source tree and setup the venv. After that all you need to do is setup the database and run the commands. I plan to add automation to ansible to have it setup the database and fully automate setup eventually, just wanted to get your initial thoughts on this.

AMDmi3:
I don't think it belongs here. Deployment should stay in separate repo, and this implementation actually contains too much custom configs and is tied to FreeBSD to be usable to the public. I don't use vargant to be able to properly test and maintain it here too (though I do use ansible to deploy it; well, us-ed since I've had to switch from packages to ports due to missing py35 flask packages and recent postgres support).

swills:
You can still use ansible for flask, as my ansible stuff does, even with Python 3.5. Also, the Vagrant stuff shouldn't get in the way, and it should be easy to add support for using another OS, but I'm happy to keep this stuff elsewhere.

PR #106, opened 2017-01-02
first pass at adding gem support

swills:


rules.yaml
```diff
@@ -571,6 +571,7 @@
 - { namepat: "ruby-(.*)",              setname: "ruby:$1", family: [ debian, arch, fedora, mageia, sisyphus, guix ] }
 - { namepat: "rubygem-(.*)",           setname: "ruby:$1", family: [ freebsd, fedora, opensuse, slackbuilds ] }
 - { namepat: "ruby[0-9][0-9]-(.*)",    setname: "ruby:$1", family: [ freebsd, openbsd, pkgsrc ] }
+- { namepat: "(.*)",                   setname: "ruby:$1", family: [ rubygem ] }
```
AMDmi3: Can just remove ```namepat``` and use ```setname:$0```

repos.yaml
```diff
@@ -631,6 +631,20 @@
...
+  sources:
+    - name: latest_specs.4.8
+      fetcher: File
+      parser: RubyGem
+      url: https://api.rubygems.org/latest_specs.4.8.gz
+      compression: gz
+  links:
+    - desc: RubyGem page
+      url: 'https://www.rubygems.org/gems/{name}/'
+  tags: [ all, have_testdata, demo, production, fastfetch ]
```
AMDmi3: This doesn't have testdata (would be nice to add some btw). And I'd not add it to production just yet.

repos.yaml
```diff
@@ -631,6 +631,20 @@
       url: 'https://pypi.python.org/pypi/{name}/'
   tags: [ all, preproduction, fastfetch ]
 
+- name: rubygem
```
AMDmi3: ```rubygems``` perhaps, to match repo name

repology/parser/rubygem.py
```diff
@@ -0,0 +1,49 @@
...
+# repology is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with repology.  If not, see <http://www.gnu.org/licenses/>.
+
+import types
+from rubymarshal.reader import loads, load
```
AMDmi3: I'd prefer ```import rubymarshal.reader``` and using fully-qualified names, otherwise it's not clear what ```load``` actually does.

rules.yaml
swills: Ah, good to know, will fix that, thanks!

repos.yaml
swills: Ok, will fix.

repos.yaml
swills: Ok, will try to add testdata. As for production, just wanted it to show up, but it adds > 100k packages so maybe something not to turn on by default.

repology/parser/rubygem.py
swills: Gotcha, will do, thanks for your comments.

repos.yaml
AMDmi3: Just add ```shadow: true``` for it as it's done for CPAN. This way the packages which are only present in this repo **only** won't show up in the metapackages index. If they are present in some other repo, they will show up. A nice way to hide packages we can't compare to anything.

swills:
Name and version are going to be the only things available without making 1 request per module...

AMDmi3:
Ok, looks almost good (just need to add ```pip install rubymarshal``` to .travis.conf and probably simplify the ruleset, I'll do it myself). I've also added py-rubymarshal to FreeBSD ports.

AMDmi3:
Oh, you've already fixed travis. Even better.

rules.yaml
```diff
@@ -572,7 +572,10 @@
 # Ruby
 - { namepat: "ruby-(.*)",              setname: "ruby:$1", family: [ debian, arch, fedora, mageia, sisyphus, guix ] }
 - { namepat: "rubygem-(.*)",           setname: "ruby:$1", family: [ freebsd, fedora, opensuse, slackbuilds ] }
+- { namepat: "rubygem-(.*)[0-9]",      setname: "ruby:$1", family: [ freebsd ] }
```
AMDmi3: These two rules do nothing, since names are already processed by `rubygem-(.*)` rule. Let's not treat gems ending with number specially for now.

AMDmi3:
Ok, looks good. Please remove two unneeded rules, rebase and squish it into a single commit to keep the history clean, and it can be merged.

AMDmi3:
Thank you!

PR #125, opened 2017-01-12
Rules: Match the correct gnupg package for Gentoo

Whissi:
Gentoo GNU Linux has two packages named "gnupg":

 - app-crypt/gnupg
 - app-vim/gnupg

Repology is currently comparing against app-vim/gnupg, see http://repology.org/metapackage/gnupg/versions

This PR tries to match the correct package. Hope that's working like intended following your documentation. Haven't tried with real data.

### DIFF
--- a/rules.yaml
+++ b/rules.yaml
@@ -391,8 +391,9 @@
 - { namepat: "gnet[0-9]+",             setname: gnet }
 - { name: gnet, verpat: "1\\..*",      setname: gnet1 }
 
-- { namepat: "gnupg[0-9]+",            setname: gnupg }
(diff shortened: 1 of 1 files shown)


AMDmi3:
Unfortunately your fix is not correct. You try to changing name from ```gnupg``` to ```gnupg``` which is essentially no-op. Instead, ```gnupg``` in category ```app-vim``` should be changed to something like ```vim-gnupg```. This also matches corresponding package in AUR (not yet shown on the site, but will be soon).

AMDmi3:
The fix will appear on the site in an hour. Thank you for submission!

PR #150, opened 2017-01-26
Gentoo Parser: Set default maintainer if no maintainer is set

Whissi:
If a package has no maintainer set in metadata.xml we use the generic
"maintainer-needed@gentoo.org" handler.


Whissi:
Wait, I'll add test coverage.

AMDmi3:
Excellent work, thank you!

PR #151, opened 2017-01-28
Rules update

Whissi:
This PR will fix version tracking for some packages.

### DIFF
--- a/rules.yaml
+++ b/rules.yaml
@@ -271,6 +271,8 @@
 - { name: dosbox,                      verpat: "0\\.74\\.r.*",       family: [ sisyphus ], ignorever: true }
 - { name: inkscape,                    ver: "0.92",                  family: [ fedora ],   ignorever: true } # prerepease
 - { name: libfaketime,                 verpat: ".*201[67].*",                                ignorever: true } # FreeBSD is using git snapshot
+- { name: kbuild,                      verlonger: 3,                                         ignorever: true } # 0.1.9998.8.2814.25 @ pkgsrc
+- { name: kbuild,                      verpat: ".*[rs].*",                                   ignorever: true } # 0.1.9998.r2813 | 0.1.9998svn2814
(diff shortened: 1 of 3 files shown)


rules.yaml
```diff
@@ -280,6 +280,7 @@
 - { name: openxcom,                    verpat: "2016.*",                                     ignorever: true } # git snapshots
 - { name: p5-catalyst-plugin-authentication, ver: "0.100230",                                ignorever: true }
 - { name: p5-device-serialport,        ver: "1.040000",                                      ignorever: true }
+- { name: tt-rss,                      verpat: "r|201[67]",                                  ignorever: true } # Kind of snapshot release
```
AMDmi3: Pattern is broken (needs ```.*```s). And I don't think we can ignore it cause

> Tiny Tiny RSS uses a rolling release model based on git master branch which is considered stable

Let's postpone this until we can change versions with rules, so we could transform dates to a single format and ignore other versions.

rules.yaml
```diff
@@ -275,6 +275,7 @@
 - { name: libfaketime,                 verpat: ".*201[67].*",                                ignorever: true } # FreeBSD is using git snapshot
 - { name: kbuild,                      verlonger: 3,                                         ignorever: true } # 0.1.9998.8.2814.25 @ pkgsrc
 - { name: kbuild,                      verpat: ".*[rs].*",                                   ignorever: true } # 0.1.9998.r2813 | 0.1.9998svn2814
+- { name: ksh,                         ver: "20120801",                  family: [ fedora ], ignorever: true }
```
AMDmi3: Let's postpone this. More repos not available on repology.org yet contain weird versions for ksh, so this will need more consideration.

AMDmi3:
Closing as it was merged, except for two items. For ```tt-rss``` I've added ignore for all dates, as it has proper upstream versioning. And for ```ksh``` there's no sense adding any rules at there's no clean upstream version. I couldn't even find its source actually.

PR #168, opened 2017-02-16
rules: map eclipse-java to eclipse on slackbuilds

aclemons:


AMDmi3:
I've just removed family restriction. Thanks for submission!

PR #175, opened 2017-03-03
add kde neon

apachelogger:


AMDmi3:
There are other repos, such as http://archive.neon.kde.org/dev/unstable/dists/xenial/main/source/Sources.gz. Which is the latest?

apachelogger:
`/user/` contains the latest builds of released software.

The `/dev/` repositories are from-git builds, so they probably aren't terribly interesting to look at from a repology point of view as they almost completely rotate once a week.

AMDmi3:
Got it, thank you!

PR #190, opened 2017-03-21
Two more python rules.

geier:
khal and todoman are python packages as well.


AMDmi3:
Thanks, will be active in ~2 hours

PR #202, opened 2017-04-05
Add openSUSE Leap 42.2 repositories

antlarr:


AMDmi3:
Sorry, but we only support source repositories. Binary repos are polluted by subpackages (xxx, xxx-32bit, xxx-devel and so forth), which cannot be counted and compared to other distros in a consistent way.

I've browsed through download.opensuse.org, though, and maybe

* http://download.opensuse.org/source/distribution/leap/42.2/repo/oss/suse/
* http://download.opensuse.org/source/distribution/leap/42.2/repo/non-oss/suse/

could be used instead? Is adding support for 42.1/42.3 as well a good idea?

antlarr:
Ah, I didn't know that you used source repositories, yes those would be the right urls. I'll change that.

42.3 is the next release and hasn't been released yet . Also 42.1 will end its maintenance lifetime on 17th of May, so I don't think it's worth it.

Btw, is it ok to use name: oss and subrepo: "Main Repository (OSS)" (as a description) ? I wasn't too sure about that.

AMDmi3:
> 42.3 is the next release and hasn't been released yet

I'd add it just to be ready for the release. It may also help during release process, but I'm not sure.

> Btw, is it ok to use name: oss and subrepo: "Main Repository (OSS)" (as a description) ? I wasn't too sure about that.

I'd go with just `oss' and `non-oss'. These need not be too descriptive and making them too long will in fact break the site design :) (see them near repo names in https://repology.org/metapackage/firefox/packages, https://repology.org/metapackage/firefox/versions)

antlarr:
42.3 will be released by the end of year, and it's just a matter of replacing all instances of "42.2" to "42.3" in repos.yaml, so I guess it can be done when it's actually released not to confuse users in the following months :)

AMDmi3:
I haven't checked yet, but won't http://download.opensuse.org/update/* contain the same binary packages?

repos.yaml
```diff
@@ -961,6 +961,43 @@
     - desc: Package details
       url: 'https://software.opensuse.org/package/{name}'
   tags: [ all, production, opensuse, rpm ]
+  #
```
AMDmi3: Please fix, also please move the block before tumbleweed to make it consistent with the freshness and sorting on the site.

antlarr:
http://download.opensuse.org/source/distribution/* contains the packages that were released as the 42.2 release, while http://download.opensuse.org/update/* contains updates to those packages.

AMDmi3:
http://download.opensuse.org/update/* contain updates to binary packages, which are binary packages themselves. To speed this up, I've added all OpenSUSE Leap repos, hope it's OK.

PR #204, opened 2017-04-07
Add rules for slackbuilds repo

bosth:
Set SlackBuilds `development/Sphinx` to `python:sphinx` and `ssr` to `simplescreenrecorder`.


AMDmi3:
Sorry, rules were split into multiple files, also rules should resolve ambiguities in all repos at once. These should go to 15.renames-ambiguities.yaml and should rename all cases of "swift" and "ssr". There should also be a fallback catch-all rule for future packages, see `mog` or `polly` as example

AMDmi3:
Both rules are already present in the ruleset, so this may be closed

PR #214, opened 2017-04-23
Improve slackbuilds download URL parsing

audriusk:
SlackBuilds repository .info files use `DOWNLOAD` and `DOWNLOAD_x86_64` to point to source tarballs according to the following rules:

- If the tarball is the same for all architectures, `DOWNLOAD_x86_64` is left empty.
- If the tarball differs for each arch, `DOWNLOAD` and `DOWNLOAD_x86_64` point to different URLs.
- If the application does not build on one of the supported architectures, the corresponding field is marked as `UNSUPPORTED`.
- If the application does not build on a clean Slackware64 system, but would if there were 32bit compatibility libraries present at build time, `DOWNLOAD_x86_64` is marked as `UNTESTED`.

Changed .info file parsing to adhere to these rules more accurately. Also fixed parsing of empty fields (e.g. `FOOBAR=""`).


AMDmi3:
Great PR, thank you! :+1::+1::+1:

PR #215, opened 2017-04-23
Improvements for Antergos Repos

lots0logs:


AMDmi3:
Um, these look like separate repos.

lots0logs:
Not sure what you mean?

AMDmi3:
These contain intersecting sets of packages, so they are not subrepositories like e.g debian main/contrib/non-free, but separate repos like debian stable/experimental are. They should not be merged.

lots0logs:
I didn't realize that was the expectation. It just seems like it makes more sense to have them together because there is really only one repo. The staging repo is just that..staging..no package stays there for longer than a few days usually. Maybe its better to remove staging altogether?

AMDmi3:
> It just seems like it makes more sense to have them together because there is really only one repo. The staging repo is just that..staging..no package stays there for longer than a few days usually

It's the same as with Debian Experimental. It is a separate repo, you can compare package versions between it and main repo, you can watch package move from staging to main, users (I assume) can enable main, or main+staging if they are eager to test freshest versions.

> Maybe its better to remove staging altogether?

No, staging is actually the most valuable for repology as newest versions appear there first.

So let's leave both repos there, I'll gladly merge link-related improvements.

lots0logs:
@AMDmi3 Done :white_check_mark:

PR #227, opened 2017-05-07
Implement generic "no-maintainer" pseudomaintainers for all repos

l2dy:


AMDmi3:
No, unlike openmaintainer which is just a flag, a list of (all, outdated, whatever) ports and with no maintainer is useful.

l2dy:
IMHO, listing ports without a maintainer should not be implemented in the parser and rely on a pseudo mail address. Other repositories don't have such conventions so if you consider this list useful, you should make the functionality repository-independent.

AMDmi3:
Actually, most repos do have these conventions, and above that, their "no-maintainer" entities have special meaning such as being a mail list. There's a lot of maillists for that purpose in Debian/Ubuntu, and other repos have maintainer-needed@gentoo.org, ports@freebsd.org, ports@openbsd.org, pkgsrc-users@netbsd.org etc. 

It's true that some repos do lack these, but it seems to me that the best way to add no-maintainer for these is through introducing a set of pseudo-emails, because
- it doesn't require code complication (extra conditions and queries for packages with no maintainer)
- it will get along nicely with repos which do have no-maintainer entities (will look and work the same way)

And in any case, if upstream has explicit no-maintainer flag, it seems to be the best to use it.

I'm turning this issue into one for implementing generic no-maintainer entry for all repos.

PS. Note to self: we'll probably need special treatment for
- repos for which we can't parse out maintainers (yet)
- repos which don't have concept of maintainer at all

l2dy:
@AMDmi3 This is a pull request so you might want to create a new issue for this.

AMDmi3:
Oh, true, I've missed that.

PR #242, opened 2017-05-15
perl:yaml-libyaml hasn't released 0.71 yet

pink-mist:
ALT Sisyphus got an unauthorised release by a 3rd party and used that
instead.

### DIFF
--- a/rules.d/20.verignores.yaml
+++ b/rules.d/20.verignores.yaml
@@ -79,6 +79,7 @@
 - { name: p5-device-serialport,        ver: "1.040000",                                    ignorever: true }
 - { name: p5-net-whois-ripe,           ver: "2.005005",                                    ignorever: true } # < 2.006
 - { name: powder-toy,                  verlonger: 2,                                       ignorever: true } # third component is a build number, it's not a part of official version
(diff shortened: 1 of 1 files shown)


AMDmi3:
This is questionable. http://www.cpan.org/authors/id/R/RU/RURBAN/ contains tarballs for `YAML-LibYAML-0.71.tar.gz` and even `YAML-LibYAML-0.73.tar.gz`. I think CPAN maintainer should be contacted instead and asked why there are tarballs for new versions, but no version on CPAN and no tags in (git repo)[https://github.com/ingydotnet/yaml-libyaml-pm].

Cast @ingydotnet

pink-mist:
The RURBAN release isn't authorised is what I'm saying. And since that release, there was an actual authorised 0.64 released by TINITA on 2017-04-07, compared to the RURBAN unauthorised release on 2016-09-08 and 2016-09-13.

pink-mist:
Compare the official release on https://metacpan.org/release/YAML-LibYAML with https://metacpan.org/release/RURBAN/YAML-LibYAML-0.71 and https://metacpan.org/release/TINITA/YAML-LibYAML-0.64

AMDmi3:
Ok, sounds reasonable.`name` should be `perl-yaml-libyaml` though, as transfotmarion to `perl:` is applied after verignore block.

pink-mist:
Alright, I changed the name :)

PR #253, opened 2017-05-25
Add Slackware support.

idlemoor:
Hi, this has no rename rules yet, and no tests, but is it ok as a starting point?
(it seems to work running locally)

### DIFF
--- a/rules.d/61.global-perl.yaml
+++ b/rules.d/61.global-perl.yaml
@@ -11,7 +11,7 @@
 - { category: dev-perl,                setname: "perl:$0", family: [ gentoo ] }
 - { category: perl-core,               setname: "perl:$0", family: [ gentoo ] }
   # Gentoo and Mageia shamelessly mangle perl module versions, ignore
-- { namepat: "perl:(.*)",              family: [ gentoo, mageia, rosa, macports ], ignorever: true }
+- { namepat: "perl:(.*)",              family: [ gentoo, mageia, rosa, macports, pclinuxos ], ignorever: true }
 
 # manual perl modules
 - { name: backuppc-xs,                 setname: "perl:backuppc-xs" }


idlemoor:
Apparently not :( I'll fix that.... cheers

repology/parser/slackware.py
```diff
@@ -0,0 +1,96 @@
...
+        #   extra/       (but NOT aspell-word-lists)
+        # Everything else is ignored:
+        #   isolinux/, kernels/, pasture/, source/, testing/,
+        #   usb-and-pxe-installers/
+
+        with open(path, 'r') as filelisttxt:
+            for line in filelisttxt.readlines():
+                fields = line.strip().split('/')
+                if re.search('\.t.z$', fields[-1]):
+                    revsplit = fields[-1][::-1].split('-', 3)
```
AMDmi3: Using `rsplit` should simplify it as long as following lines.

repology/parser/slackware.py
```diff
@@ -0,0 +1,96 @@
...
+                        if pkgnam in packages:
+                            if packages[pkgnam].category == 'extra':
+                                # overwrite the existing instance from extra/ (e.g. nouveau blacklist)
+                                packages[pkgnam].category = fields[2]
+                                packages[pkgnam].version = version
+                            else:
+                                # safe to assume the existing instance came from patches/, do nothing
+                                pass
+                        else:
+                            pkg = Package()
```
AMDmi3: I'd prefer duplication of these lines (pkg, pkg.name, pkg.version, pkg.maintainers, packages[]) reduced. You can create package and set name/version/maintainers and add it into the dict unconditionally.

repology/parser/slackware.py
```diff
@@ -0,0 +1,96 @@
...
+            for line in filelisttxt.readlines():
+                fields = line.strip().split('/')
+                if re.search('\.t.z$', fields[-1]):
+                    revsplit = fields[-1][::-1].split('-', 3)
+                    if len(revsplit) != 4:
+                        continue
+                    pkgnam = revsplit[3][::-1]
+                    version = revsplit[2][::-1]
+
+                    if fields[1][:9] == 'slackware' and fields[2] != 'kdei':
```
AMDmi3: I don't see `slackware*` there, only `slackware`, so this could be simplified. Why exccude kdei? These may be merged into a single package kde-l10n or kde-i18n like we do with other repos.

repology/parser/slackware.py
```diff
@@ -0,0 +1,96 @@
...
+        #   isolinux/, kernels/, pasture/, source/, testing/,
+        #   usb-and-pxe-installers/
+
+        with open(path, 'r') as filelisttxt:
+            for line in filelisttxt.readlines():
+                fields = line.strip().split('/')
+                if re.search('\.t.z$', fields[-1]):
+                    revsplit = fields[-1][::-1].split('-', 3)
+                    if len(revsplit) != 4:
+                        continue
```
AMDmi3: Prefer print here. If there are non-conforming entries, either data or a parsing algorithm is broken, it's better to detect either case.

repology/parser/slackware.py
```diff
@@ -0,0 +1,96 @@
...
+
+class SlackwareParser():
+    def __init__(self):
+        pass
+
+    def Parse(self, path):
+        packages = {}
+
+        # For all the Slackware packages:
+        maintainers = ['volkerdi@slackware.com']
```
AMDmi3: Hmm. Not sure about it, but let it be, since sometime we'll generate default maintainer anyway.

repology/parser/slackware.py
```diff
@@ -0,0 +1,96 @@
...
+                            pass
+                        else:
+                            pkg = Package()
+                            pkg.category = 'extra'
+                            pkg.name = pkgnam
+                            pkg.version = version
+                            pkg.maintainers = maintainers
+                            packages[pkg.name] = pkg
+
+        return [package for package in packages.values()]
```
AMDmi3: Just `packages.values()`. Or `list(packages.values())`

repology/parser/slackware.py
```diff
@@ -0,0 +1,96 @@
...
+                            packages[pkgnam].version = version
+                        else:
+                            pkg = Package()
+                            pkg.category = 'patches'
+                            pkg.name = pkgnam
+                            pkg.version = version
+                            pkg.maintainers = maintainers
+                            packages[pkg.name] = pkg
+
+                    elif fields[1] == 'extra' and fields[2] != 'aspell-word-lists':
```
AMDmi3: Also don't like to exclude specific port. Can be handled with a rule.

repos.d/slackware.yaml
```diff
@@ -0,0 +1,48 @@
+###########################################################################
+# Slackware
+###########################################################################
+- name: slackware-current
```
AMDmi3: I'm using underscores in internal repo names, let's stick to this "rule" here as well.

README.md
```diff
@@ -70,6 +70,7 @@ version are always parsed.
 | PyPi                             | ✔       |         |       | ✔ (2) |         |          |
 | RubyGems                         |         |         |       | ✔ (2) |         |          |
 | SlackBuilds                      |         | ✔       | ✔     | ✔     |         | ✔        |
+| Slackware                        |         | ✔       | ✔     | ✔     |         |          |
```
AMDmi3: But you don't really parse WWW.

PR #275, opened 2017-07-17
Update running and rules docs.

idlemoor:


docs/RUNNING.md
```diff
@@ -65,7 +65,7 @@ First, let's try to fetch some repository data and see if it works.
 No database is needed at this point.
 
 ```
-./repology-update --fetch --parse
```
AMDmi3: This change is not correct. The idea is that ```family``` condition correspond to ```family``` field from repository definitions. schemacheck is a derivative from that.

AMDmi3:
Merged, thank you!

PR #283, opened 2017-07-23
Allow archive.org/.is for discontinued homepages.

idlemoor:
In repology/database.py the discontinued sites googlecode, codeplex and gna are recognised by simple SQL wildcards.  Both archive.org and archive.is have good coverage of googlecode, codeplex and gna. Unfortunately, when the dead homepages are replaced by archive.org, the new url includes the old url, and so the SQL wildcard still matches, and so the new URL is incorrectly listed as a problem.

Example: http://web.archive.org/web/20170206035443/https://gna.org/projects/ngjackspa

This fix simply excludes urls that contain archive.org or archive.is


AMDmi3:
Good catch, thank you! I've redone it in a bit more correct way.

PR #290, opened 2017-07-29
Add rule for cpanminus

pink-mist:


AMDmi3:
Thanks, will be deployed on Monday

Issue #268, opened 2017-07-08
Merge webkit metapackages


jbicha:
Test Case
=======
I'd like to see what version of the latest [webkitgtk](https://webkitgtk.org/) is offered in different distros. But webkitgtk packaging is particularly unique in its naming.

- [libwebkit-gtk](https://dev.solus-project.com/source/libwebkit-gtk/)  - Solus (not yet in repology)
- [webkit2gtk](https://repology.org/metapackage/webkit2gtk/versions) -  Arch Linux and Debian
- [webkit2gtk3](https://repology.org/metapackage/webkit2gtk3/versions) - OpenSUSE
- [webkitgtk4](https://repology.org/metapackage/webkitgtk4/versions) - Fedora

I'd like to see those all linked together or matched so that it's easy to see. Otherwise, it requires manually figuring out the source package names for all the different distros and loading those pages manually.

I'm proposing adding a bit of metadata so that Repology understands that all those different source packages are packaging the same source.

AMDmi3:
Repology does that from the start, see rules.d directory in the repo. There are no rules for webkit yet (although I've already sern requests to merge it) because I'm too confused by the number of webkit packages

https://repology.org/metapackages/all/?search=webkit

so I'd be grateful for some insight here: are all these webkit*gtk* a single project (some repos just using outdated versions), or are there multiple branches developed in parallel, or are there multiple projects from different authors? As result, into how many metapackages should webkit stuff be merged?

pink-mist:
It's very confusing indeed. As far as I understand it, there are two different APIs for **version 1** and **version 2** of webkitgtk, and while **version 1** has been deprecated, a lot of software still uses it, so having a different package for it is still useful.

 * **Version 1** is usually just named ```webkitgtk``` (or ```libwebkit-gtk```) without a version between webkit and gtk, and
* **version 2** is usually named ```webkit2gtk```

Then the version for ```gtk``` can also be specified, for example the SlackBuilds.org repository has these: ```webkitgtk```, ```webkitgtk3```, and ```webkit2gtk```, where the first two are for **version 1**, differing in which version of gtk they're built for (they're built for ```gtk 2``` and ```gtk 3``` respectively) and the last one is for **version 2**.

jbicha:
@AMDmi3 all of the ones I specified in my original comment represent the same library.

[...]

PR #291, opened 2017-07-29
Add rules for webkit

pink-mist:
See #268 

I'm not entirely certain if this has any unintended matches.

There's also the following ones which I did not add rules for, and I'm unsure if they should be included or not:
* https://repology.org/metapackage/mingw:webkitgtk/versions
* https://repology.org/metapackage/mingw:webkitgtk3/versions

### DIFF
--- a/rules.d/45.splitmergever.yaml
+++ b/rules.d/45.splitmergever.yaml
@@ -579,6 +579,11 @@
 - { namepat: "vte[0-9._-]+",           setname: vte }
 - { name: [vte-common,vte-gtk2-compat,vte-gtk2-xfce,vte-gtk3,vte-reference], setname: vte }
 
+- { namepat: "libwebkitgtk[234]?",     setname: webkitgtk }
+- { namepat: "webkit(24?)?(-?gtk)+[1234]?(-2[.]0)?", setname: webkitgtk }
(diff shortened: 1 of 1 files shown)


AMDmi3:
Please put it into 45.splitmergever, also it's better to not use . (.?, .*) as it may have unintrnded matches (for example, something like webkitgtk-python). It's common practice to use limited regexp which matches version numbers and srparators, e.g. something like `(?:lib)?webkit[0-9._-]*gtk[0-9._-]*` plus additional rule for unstable with ignorever.

No, mingw32 may be ignored for now. The idea is to merge them all in one go as soon as flavors are implemented (so they compare properly against native packages).

pink-mist:
Alright, I moved it over to 45 and reworked the pattern.

pink-mist:
And now I also split out the -unstable one and set it to ignorever: true

pink-mist:
Also added rules for qtwebkit

AMDmi3:
Please simplify. `[0-9._-]*` may be used instead of any version-like sequence

AMDmi3:
To speed this up, I've just added the rules. For qtwebkit we already have a common rule for qt packages webkit nicely fits into.

PR #300, opened 2017-08-10
Recognise openssh-.* as openssh

n-st:
Examples:
- openssh-portable on FreeBSD
- openssh-{client-utils,client,keygen,moduli,server-pam,server,sftp-avahi-service,sftp-client,sftp-server} on OpenWRT

### DIFF
--- a/rules.d/10.renames-misc.yaml
+++ b/rules.d/10.renames-misc.yaml
@@ -600,6 +600,7 @@
 - { name: opencv2,                     setname: opencv                    }
 - { name: [openjp2, libopenjpeg2.0],   setname: openjpeg2                 }
 - { name: openshot-qt,                 setname: openshot }
+- { name: openssh-.*,                  setname: openssh                   }
 - { name: [openvpn-mbedtls, openvpn-polarssl, openvpn2, openvpn23], setname: openvpn }
 - { name: open-tyrian,                 setname: opentyrian }
 - { name: open-babel,                  setname: openbabel }


AMDmi3:
The fix is not really correct:
- It should've been "namepat", not "name"
- You can't use pattern here as it matches unrelated projects (openssh-askpass*, openssh-blacklist)

I've added rules for chosen openssh-* packages in 7b757120e65f0856a08daba62cb02c83a0b1fa20, though I don't see any openssh* packages from either OpenWRT or LEDE. Am I missing some repositories?

n-st:
Ah, sorry for the mis-fix and thanks for correcting that.

> I don't see any openssh* packages from either OpenWRT or LEDE

I didn't check where you're fetching the OpenWRT package list from; I just got those package names from http://downloads.openwrt.org/snapshots/trunk/ar71xx/generic/packages/packages/. Just noticed there's also https://github.com/openwrt/packages/blob/master/net/openssh/Makefile.
Are those two any good?

Got to say, I don't have an active OpenWRT system at the moment (nor one running Void Linux, see PR #301); I'm just curious how up-to-date they are. ;)

n-st:
Oh, and I think you've got a copy-paste error in 7b757120e65f0856a08daba62cb02c83a0b1fa20. I added an [inline comment](https://github.com/repology/repology/commit/7b757120e65f0856a08daba62cb02c83a0b1fa20##commitcomment-23746436) about that.

AMDmi3:
Copypasta was fixed, and I know a cause for missing openssh packages for openwrt: it's set up as a shadow repository, which means that packages from OpenWrt/LEDE which are not present in other repos are hidden - the intent was to hide packages specific to embedded systems which cannot be compared to any other repos. Maybe these should be unhidden and/or OpenWrt/LEDE parsing tseaked to use `Source` field:

```
Package: openssh-client-utils
Version: 7.5p1-4
Depends: libc, libopenssl, zlib, openssh-client, openssh-keygen
Source: feeds/packages/net/openssh
License: BSD ISC
LicenseFiles: LICENCE
Section: net
Maintainer: Peter Wagner <tripolar@gmx.at>
Architecture: ar71xx
Installed-Size: 634797
Filename: openssh-client-utils_7.5p1-4_ar71xx.ipk
Size: 633446
MD5Sum: 887309c64cf1ae21d8ffbb73ecae3efb
SHA256sum: e844af7dd0b9e13e037a64573154d7b63482c57a44070bb8dd40c39878c98c1b
Description:  OpenSSH client utilities.
```

which will turn a lot of split packages like `openssh-*` into comparable biggerr ones like `openssh`.

AMDmi3:
Superseded by #306

PR #301, opened 2017-08-10
Parser for the Void Linux distribution

n-st:
Mostly copied from the GoboLinux parser, with some changes for variable replacements in package's distfile URLs.

### DIFF
--- a/repology/parser/parsers.py
+++ b/repology/parser/parsers.py
@@ -50,4 +50,5 @@ from .slackbuilds import SlackBuildsParser
 from .snap import SnapJsonParser
 from .spec import SpecParser
 from .srclist import SrcListParser
+from .voidlinux import VoidLinuxGitParser
 from .yacp import YACPGitParser
--- a/repology/parser/voidlinux.py
+++ b/repology/parser/voidlinux.py
@@ -1,0 +1,105 @@
+# Copyright (C) 2016-2017 Dmitry Marakasov <amdmi3@amdmi3.ru>
+#
+# This file is part of repology
+#
+# repology is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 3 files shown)


AMDmi3:
How many packages skipped because of unknown variables there?

I don't really like this approach. This is shell, and with shell any construct is possible, including using custom variables, math operations, complex quotes etc. I'm seriously thinking on ripping a parser out of bash to make a library which generates shell AST, which can be used to selectively reliably parse shell-based build scripts.

n-st:
I'm wary of shell script parsing too, but I figured the primitive approach is better than not having the repo at all.

Currently, 853 packages use variables in their `distfiles` attribute that couldn't be parsed. Note, however, that this only affects the `downloads` attribute.

I only just realised that packages could theoretically also use variables in their other attributes, even `name` and `version` — and indeed, there are currently 58 packages that do that (and another 71 packages with variables in their `description`). That's 1.2% of the 4790 packages total.  
The parser becomes ever less useful…

Still, I can't think of a way to circumvent this, short of re-inventing shell scripting inside repology…

AMDmi3:
Either implementing a shell parser, or making upstream publish preprocessed data, yes.

n-st:
I don't really see either of that happening anytime soon…
What do you think about including the imperfect parser until then ("better than nothing" etc)?

AMDmi3:
> I don't really see either of that happening anytime soon…

While writing custom shell parser really requires a lot of effort, making a distro provide metadata for its packages should be quite doable. I suspect a distro already has this data available, it should just publish it in a machine readable format.

> What do you think about including the imperfect parser until then ("better than nothing" etc)?

I'd strongly prefer not to implement any partially supported repositories. It leads to incorrect statistics, incomplete package lists for maintainers, general confusion and disappointment.

Issue #230, opened 2017-05-10
Implement generic "no-maintainer" pseudomaintainer for all repos


AMDmi3:
Of repos which we can parse maintainers from, some repos do have a special "no-maintainer" entry like maintainer-needed@gentoo.org, ports@freebsd.org, ports@openbsd.org, pkgsrc-users@netbsd.org etc., while some don't. To support unmaintained port lists for the latter in a generic way, we could automatically generate pseudomaintainer entries for ports with no maintainers in a generic way (e.g. in normalization phase). `no-maintainer-<repo>@repology` or something like that. We also need to treat repos in which there are no maintainers or for which we can't parse maintainers yet in a special way here.

See also #227

AMDmi3:
Additional thought: there should be a way to specify default maintainer via [repository configs](https://github.com/repology/repology/tree/master/repos.d). This will remove default maintainers from the code which makes it more generic. If there isn't default maintainer specified, synthetic value should be used.

PR #302, opened 2017-08-12
Only use the Gentoo default maintainer for the main Gentoo repo.

kensington:


AMDmi3:
Sorry, but I don't really like this. Using `repo_name` would make sense if it was used for something generic, but if it's only used to differentiate main gentoo repo from overlays it doesn't feel right, especially while there are more straightforward ways to do it.

Let's instead fix it in a generic way in issue #230.

kensington:
OK, thanks for the feedback. I'll try and propose some patches for issue #230 at some point.

AMDmi3:
Don't bother, I have a solution ready

Issue #294, opened 2017-08-06
ravenports: new php extension ports aren't matching with existing metapackages


jrmarino:
see: https://repology.org/metapackages/unique-in-repo/ravenports/

All new php71 extension ports are currently considered "unique" despite having identical names as FreeBSD ports collection.

jrmarino:
I guess the most effective approach in the future would be to create pull requests with rule updates that would just be approved for merge?  

Is anybody doing that yet and is that what you prefer?

AMDmi3:
> I guess the most effective approach in the future would be to create pull requests with rule updates that would just be approved for merge?

Yes, as log as they are simple and correct, e.g. don't require a lot of iterations to fix.

PR #307, opened 2017-08-25
Support php on ravenports (issue #294)

jrmarino:
This single line should be to solution to issue #294.


jrmarino:
I see some activity on repology. it would be nice to get this one in today, it affects like 70 metapackages, so it's a big one (and will grow to ~200 packages with php56 and php72 coming)

PR #311, opened 2017-09-01
Merge rabbitmq-server and rabbitmq

l2dy:
https://repology.org/metapackage/rabbitmq-server/versions
https://repology.org/metapackage/rabbitmq/versions


AMDmi3:
(will deploy on Monday)

PR #315, opened 2017-09-03
Complete merging of ravenports uniques (existing issue discovered)

jrmarino:
Whenever possible, the rules I create here address all other repositories with similar orphaned metapackages.  for example, rather than just fix "xorg-xkeyboard-config" for ravenports, I created a rule that merged 5 orphaned xkeyboard-config ports.

Everything was straightforward except fontutil.  I detailed the issue in the commit message, but the synopsis is that a too-broad rule in section 88 converted xorg's font-util to "fonts:util" which is probably undesired.  The rule I added in section 16 merges *numerous* orphans to the existing "fonts:util" metapackage so at least they are all together, but you probably want to fix the original rule and move these to the "font-util" metapackage.

### DIFF
--- a/rules.d/10.renames-misc.yaml
+++ b/rules.d/10.renames-misc.yaml
@@ -795,6 +795,7 @@
 - { name: valyria-tear,                setname: valyriatear               }
 - { name: [varnish4,varnish5],         setname: varnish                   }
 - { name: vdr-devel,                   setname: vdr, ignorever: true      }
+- { name: video4linux,                 setname: v4l-compat                }
 - { name: [vim-console,vim-core,vim-full,vim-gnome,vim-gtk2,vim-gtk3,vim-light,vim-lite,vim-minimal,vim-motif,vim-runtime,vim-share,vim-x11,vim-xaw], setname: vim }
 - { name: vlc-nightly,                 setname: vlc, ignorever: true      }
 - { name: [vlc-qt,vlc-qt4,vlc-qt5,vlc-nox,vlc-decklink], setname: vlc, ignorever: true }
--- a/rules.d/12.renames-fonts.yaml
+++ b/rules.d/12.renames-fonts.yaml
@@ -577,6 +577,8 @@
     - xorg-font-micro-misc
   setname: "fonts:micro-misc"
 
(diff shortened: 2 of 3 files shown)


jrmarino:
Hi Dmitri,
What is your plan for this one?  It won't apply anymore.  Do you want me to separate them all into individual patches and mail them or something? (which can also be do directly on github with the ".patch" suffix on the url).
I literally spent hours on this and tried to help as many repos as I could, so I would hate to see that contribution unused.  

I can't really do any more contributions until this one is dispositioned because I have to sync up the repo.
I was also wary of sending a group of patches because if you didn't like one, maybe the whole bunch would be rejected ...

AMDmi3:
It was broken to begin with, see travis. And yes, I'd prefer this as a bunch of separate PRs, because some things can be merged as is and some cannot, and it'd be more convenient to handle them in parallel. At first I've planned to cherry pick good commits from this, but you know, extra handwork automatically lowers the priority of a PR. I've already fixed video4linux btw.

jrmarino:
i apology for an apparent syntax error.  It's very difficult to set up a working repology so I have no way to test it (that I know of).
That said, you take about a week to accept a single PR.  So this represents about 8 weeks worth of submitting then waiting then submitting next one.

For me to continue to be motivated, It would be nice if you could pick each commit individually and apply (assuming most are syntax-good).  Whatever left I can resubmit.

I just know that if I had somebody contributing to ravenports like I'm trying to contribute to repology, I'd do everything possible to facilitate it to encourage more.

Anyway, please cherry pick the good ones.  At the very least, I've done the research which took the most time, so even that should be useful to you.

AMDmi3:
> i apology for an apparent syntax error

The testing process is underdocumented, but `make test` would check YAML schema and there's usiful sideeffect of that `./repology-update.py` without args will do nothing, but load rules, thus compile regexps thus check them. It doesn't require anything (e.g. database setup) apart from installing depends.

> Anyway, please cherry pick the good ones.

Sure, let's go this way. I haven't forgotten about this PR.

AMDmi3:
Everything merged. Regarding the last commit, I've ignored xorg fonts metaports instead.

jrmarino:
I appreciate that.  I also learn from seeing the mistakes corrected / alternate rules.

PR #317, opened 2017-09-04
Add OpenMandriva repositories

berolinux:
Signed-off-by: Bernhard Rosenkränzer <bero@lindev.ch>

### DIFF
--- a/repos.d/rpm/openmandriva.yaml
+++ b/repos.d/rpm/openmandriva.yaml
@@ -1,0 +1,40 @@
+###########################################################################
+# OpenMandriva
+###########################################################################
+- name: openmandriva_3
+  type: repository
+  desc: OpenMandriva 3.x
+  family: openmandriva
+  color: '045e9d'
+  sources:
+    - name: [ main/release, main/updates, main/testing, contrib/release, contrib/updates, contrib/testing, non-free/release, non-free/updates, non-free/testing, restricted/release, restricted/updates, restricted/testing ]
+      fetcher: Repodata
+      parser: Repodata
+      url: 'http://abf-downloads.openmandriva.org/3.0/repository/SRPMS/{source}/'
+      subrepo: '{source}'
+  repolinks:
+    - desc: OpenMandriva home
+      url: https://openmandriva.org/
+  packagelinks:
+    - desc: Package repository
+      url: 'https://github.com/OpenMandrivaAssociation/{name}'
+  tags: [ all, production, openmandriva, rpm ]
+
+- name: openmandriva_cooker
+  type: repository
+  desc: OpenMandriva Cooker
+  family: openmandriva
+  color: '045e9d'
+  sources:
+    - name: [ main/release, main/updates, main/testing, contrib/release, contrib/updates, contrib/testing, non-free/release, non-free/updates, non-free/testing, restricted/release, restricted/updates, restricted/testing ]
+      fetcher: Repodata
+      parser: Repodata
+      url: 'http://abf-downloads.openmandriva.org/cooker/repository/SRPMS/{source}/'
+      subrepo: '{source}'
+  repolinks:
+    - desc: OpenMandriva home
+      url: https://www.openmandriva.org/
+  packagelinks:
+    - desc: Package repository
+      url: 'https://github.com/OpenMandrivaAssociation/{name}'
+  tags: [ all, production, openmandriva, rpm ]
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ version are always parsed.
 | nixpkgs                          | ✔       | ✔       |       | ✔     | ✔       |          |
 | OpenBSD                          | ✔       | ✔       | ✔     |       |         |          |
 | OpenIndiana                      | ✔       |         | ✔     | ✔     |         | ✔        |
+| OpenMandriva                     | ✔       |         | ✔     | ✔     | ✔       |          |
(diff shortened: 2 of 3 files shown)


repos.d/rpm/openmandriva.yaml
```diff
@@ -0,0 +1,42 @@
+###########################################################################
+# OpenMandriva
+###########################################################################
+- name: openmandriva_3
+  type: repository
+  desc: OpenMandriva 3.x
+  family: openmandriva
```
AMDmi3: Please add this to `families` list in `repology-schemacheck.py`, this should also fix travis build

repos.d/rpm/openmandriva.yaml
```diff
@@ -0,0 +1,42 @@
+###########################################################################
+# OpenMandriva
+###########################################################################
+- name: openmandriva_3
+  type: repository
+  desc: OpenMandriva 3.x
```
AMDmi3: Is it not worth to create separate repo for each 3.x? As I see, there's only 3.0 currently, but may 3.1 also appear?

repos.d/rpm/openmandriva.yaml
```diff
@@ -0,0 +1,42 @@
...
+  family: openmandriva
+  color: '045e9d'
+  sources:
+      # XXX: */backports_testing and */updates_testing not included, are these needed?
```
AMDmi3: There are no `*/backports_testing and */updates_testing`, please remove comment.

repos.d/rpm/openmandriva.yaml
```diff
@@ -0,0 +1,42 @@
...
+  tags: [ all, production, openmandriva, rpm ]
+
+- name: openmandriva_cooker
+  type: repository
+  desc: OpenMandriva Cooker
+  family: openmandriva
+  color: '045e9d'
+  sources:
+      # XXX: */backports_testing and */updates_testing not included, are these needed?
+    - name: [ main/release, main/updates, main/testing, non-free/release, non-free/updates, non-free/testing, restricted/release, restricted/updates, restricted/testing ]
```
AMDmi3: Most of these subrepos are not there, there's only */release. Same question for contrib repo.

repos.d/rpm/openmandriva.yaml
```diff
@@ -0,0 +1,42 @@
...
+# OpenMandriva
+###########################################################################
+- name: openmandriva_3
+  type: repository
+  desc: OpenMandriva 3.x
+  family: openmandriva
+  color: '045e9d'
+  sources:
+      # XXX: */backports_testing and */updates_testing not included, are these needed?
+    - name: [ main/release, main/updates, main/testing, non-free/release, non-free/updates, non-free/testing, restricted/release, restricted/updates, restricted/testing ]
```
AMDmi3: Isn't `contrib` repo worth adding?

repos.d/rpm/openmandriva.yaml
```diff
@@ -0,0 +1,42 @@
...
+      fetcher: Repodata
+      parser: Repodata
+      url: 'http://abf-downloads.openmandriva.org/3.0/repository/SRPMS/{source}/'
+      subrepo: '{source}'
+  repolinks:
+    - desc: OpenMandriva home
+      url: http://openmandriva.org/
```
AMDmi3: https://

repos.d/rpm/openmandriva.yaml
```diff
@@ -0,0 +1,42 @@
...
+  sources:
+      # XXX: */backports_testing and */updates_testing not included, are these needed?
+    - name: [ main/release, main/updates, main/testing, non-free/release, non-free/updates, non-free/testing, restricted/release, restricted/updates, restricted/testing ]
+      fetcher: Repodata
+      parser: Repodata
+      url: 'http://abf-downloads.openmandriva.org/cooker/repository/SRPMS/{source}/'
+      subrepo: '{source}'
+  repolinks:
+    - desc: OpenMandriva home
+      url: http://www.openmandriva.org/
```
AMDmi3: https://

berolinux:
Thanks, fixed in updated patch.

For the 3.x bit, it's actually meant the way it is now. There won't be a 3.1 release, but there have been 3.01 and 3.02 releases (and there will be 3.03 and probably 3.04), but all of those are just respins of 3.0 with everything in updates/ included -- so installing 3.0 and then installing all updates is exactly the same as installing 3.02.

AMDmi3:
It turns out that `repodata` subdirectory only is only present in contrib/release and main/release for both repos. We'll only be able to parse these subrepos.

AMDmi3:
Also the 3.x and cooker repo contents seem to be completely the same.

berolinux:
That's a temporary issue with the infrastructure, we're currently in the process of switching from the old style Mandrake metadata (see e.g. http://abf-downloads.openmandriva.org/cooker/repository/SRPMS/main/release/media_info/ ) to "standard" repodata.

Both cooker metadata being out of sync and missing metadata for updates, restricted, non-free etc. should be fixed soon.

AMDmi3:
Okay, then it's good. I'll add some required changes (leave only repos which have repodata for now, add required global rules, teach repodata parser to decompress .xz etc.) and merge this. Please submit an issue/PR when more repositories are converted to repodata and may thus be added to repos.d.

AMDmi3:
It's [available](https://repology.org/statistics#openmandriva_3) on the site.

Issue #295, opened 2017-08-07
Comprehensive analysis of aspell done, report on issues and recommended fixes


jrmarino:
The aspell project is kind of a mess.  Some of this mess is causing errors in Repology.
After importing aspell packages and using Repology to detect errors, these are the ones left over that need some manual fixes (affects all platforms).

1) **Change in version scheme,  ports with lower version numbers are actually the latest version:**

aspell-el: 0.08 is newer than 0.50-3, needs rule
ref: https://ftp.gnu.org/gnu/aspell/dict/el

aspell-ca: three versions defined: 0.50.2, 2.1.5.1 (newest), 20040130.1 (needs rule)
ref: https://ftp.gnu.org/gnu/aspell/dict/ca

aspell-fo: three version defined 0.50.2, 0.51.0, 0.2.16.1 (newest)  (rule: 0.2 > 0.51)
ref: https://ftp.gnu.org/gnu/aspell/dict/fo

2) **Bogus version, taking on main "aspell" version instead of dictionary version ("0.50" or "0.60")**

aspell-be: only one version defined: 0.01
ref: https://ftp.gnu.org/gnu/aspell/dict/be
versions to ignore: "0.60", "0.50.0" [mageia/rosa]

aspell-br: two versions defined: 0.50.1 and 0.50.2
ref: https://ftp.gnu.org/gnu/aspell/dict/br
versions to ignore: "0.60.2" [mageia/rosa]

aspell-fi: only one version defined: 0.7.0
ref: https://ftp.gnu.org/gnu/aspell/dict/fi
versions to ignore: "0.60" [mageia/rosa/PCLinuxOS]

aspell-fr: only one version defined: 0.50.3
ref: http://gnu.mirror.iweb.com/aspell/dict/fr/
versions to ignore: "0.60" [mageia/gentoo/PCLinuxOS]

aspell-sl: only one version defined: 0.50
ref: https://ftp.gnu.org/gnu/aspell/dict/sl
versions to ignore: "0.60" (ubuntu/debian)
A search for independently maintained SL aspell produced nothing

aspell-sr: only one version defined: 0.02
ref: http://gnu.mirror.iweb.com/aspell/dict/sr/
versions to ignore: "0.60" (gentoo/rosa)

aspell-te: only one version defined: 0.01.2
ref: https://ftp.gnu.org/gnu/aspell/dict/te
versions to ignore: "0.60" (rosa)

aspell-tet: only one version defined: 0.1.1
[...]

PR #328, opened 2017-09-13
Finish aspell ignore work (issue #295)

jrmarino:
I tried to run "make test", but even after installing several packages, there were still some missing dependencies so I finally gave up.  It would help if there were a list of dependencies document somewhere required to make "make test" work (preferably with freebsd ports equivalents along side).  I hope there is no syntax error here.

This commit should allow issue #295 to be closed.


AMDmi3:
Good. Note that if not all packages are fixed by these rules, it's either update problem (centos is failing to update for 12 hours, fedora fails occasionally) or additional rule may be needed like for `aspell-el`.

>  It would help if there were a list of dependencies document somewhere required to make "make test" work (preferably with freebsd ports equivalents along side).

* https://github.com/repology/repology/blob/master/docs/RUNNING.md
* https://github.com/repology/freebsd-ports-repology

PR #356, opened 2017-10-11
Update broken link on metapackage report page

### LINKED ISSUE #355: Broken link to ruleset on metapackage report page
For example https://repology.org/metapackage/linux/report contains a link "ruleset" which points to https://github.com/AMDmi3/repology/blob/master/rules.yaml. This repository doesn't contain a rules.yaml file.

maxwell-k:
Fix 404 link to absent rules.yaml file by linking to the rules page in the documentation.

Fixes #355


AMDmi3:
`rules.yaml` was split into `rules.d/*.yaml`, so the link should actually point to https://github.com/repology/repology/tree/master/rules.d. I've fixed it, thank you for the report!

PR #375, opened 2017-11-11
[wip] dockerize build

avindra:
This introduces a containerized build for repology, which effectively enables you to just use  `docker-compose` to build the entire stack.

It is currently in a working state, but documentation needs to be added on how to use it.

I'm opening up the PR now in hopes of soliciting feedback from @AMDmi3 and the community using repology so far.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM python:3.6
+
+WORKDIR /usr/src/app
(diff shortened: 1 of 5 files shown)


AMDmi3:
Sorry, I won't accept unrelated stuff I can't test and maintain.

PR #377, opened 2017-11-13
Merge pyqt4/pyqt5 into python:pyqt4/pyqt5

The-Compiler:
Those are the same:

https://repology.org/metapackage/pyqt4
https://repology.org/metapackage/python:pyqt4

https://repology.org/metapackage/pyqt5
https://repology.org/metapackage/python:pyqt5

---

This is my first time contributing here and this is totally untested - so please let me know if something is wrong!

### DIFF
--- a/rules.d/70.global-python.yaml
+++ b/rules.d/70.global-python.yaml
@@ -80,6 +80,8 @@
     - pylint-plugin-utils
     - pypoppler
     - pyproj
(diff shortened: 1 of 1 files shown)


AMDmi3:
> This is my first time contributing here and this is totally untested - so please let me know if something is wrong!

It's perfect, thank you!

PR #381, opened 2017-11-14
merge dseg fonts

alexmyczko:


AMDmi3:
Thanks! I've also added a rule to fix 030 version which would otherwise be counted newer than 0.43, while it's actually older 0.30.

PR #384, opened 2017-11-14
also merge names for other hledger packages

simonmichael:
-ui, -web, -lib, -interest etc.

### DIFF
--- a/rules.d/77.global-haskell.yaml
+++ b/rules.d/77.global-haskell.yaml
@@ -20,6 +20,6 @@
 
 # APPS, not modules
 - { name: "haskell:darcs",             setname: darcs              }
-- { name: "haskell:hledger-web",       setname: hledger-web        }
+- { namepat: "haskell:hledger(?!-lib)", setname: hledger$1         }
 - { name: "haskell:xmobar",            setname: xmobar             }
 - { name: "haskell:xmonad",            setname: xmonad             }


AMDmi3:
Why not `namepat: "haskell:(hledger(?:-.+))", setname: "$1"` or just `namepat: "haskell:(hledger.*)", setname: "$1"`. Also is it correct to remove haskell: from all of them? I tend to remove languare prefixes for utilities, and at least hledger-lib is probably just the module

simonmichael:
@AMDmi3 I kept it simple to be obviously correct. Yes perhaps you're right about hledger-lib, it's the only one not providing an end-user tool. So I will probably do:
```
haskell:hledger(.*) -> $1
hledger-lib -> haskell:hledger-lib
```
? (Unless this works: `haskell:hledger(?!:-lib)` ?)

simonmichael:
Notifications may have been unclear: there's a cleaned-up commit 52fda5b here.

simonmichael:
Thanks @AMDmi3. I read that the site updates hourly but https://repology.org/metapackage/hledger/badges and https://repology.org/metapackage/haskell:hledger/badges look the same. I wonder if this is working ?

AMDmi3:
The regexp was incorrect, in fact it has broken update for other repos - it turns that (?!) lookagead construct does not generate capture, so using `$1` in setname was incorrect. I've fixed it by splitting into two rules.

simonmichael:
Thanks. And, thanks for repology, it's great!

PR #385, opened 2017-11-14
Add navigation links (ToC) to /statistics

zopieux:
image


AMDmi3:
Care to make it more semantic? E.g. something similar to how bootstrap's breadcrumbs are implemented.

PR #386, opened 2017-11-14
Make the freshness map more legible by hiding colliding labels on hover

zopieux:
image

A real fix would be to use eg. d3js to implement zooming. Please tell me if you're interested.

### DIFF
--- a/repologyapp/templates/map.svg
+++ b/repologyapp/templates/map.svg
@@ -4,6 +4,31 @@
 {% set bottomoffset = 20 %}
 
 <svg xmlns="http://www.w3.org/2000/svg" width="{{ width }}" height="{{ height }}">
+	<script type="text/javascript"><![CDATA[
+		let colliding = null;
+		let inter = (a, b) => (a.x < b.x + b.width) && (a.x + a.width > b.x) && (a.y < b.y + b.height) && (a.height + a.y > b.y);
+		let collides = (a, b) => {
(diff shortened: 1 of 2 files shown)


AMDmi3:
Wow, nice! Though I wonder if this will work on the site which has Content-Security-Policy header which prohibits builtin scripts.

arrowd:
It'd be great to have this on the main site. What's blocking this?

zopieux:
> It'd be great to have this on the main site. What's blocking this?

:shrug:

AMDmi3:
Like I've said, embedded scripts and styles are prohibited by CSP, so it won't work on the site.

zopieux:
Is your point that the embedded script should be extracted in a standalone js file instead of being embedded?

AMDmi3:
Yes, that would work.

AMDmi3:
@zopieux although, if you plan to work on this, we may consider adding an API endpoint which returns JSON with info for all these dots, and generate svg with JavaScript from the scratch instead of moving pregenerated svg objects around. This could allow to conveniently explore the while graph at once given that the points are unevenly distributed. But, I'll happily merge any solution - as said, CSP issue is the only thing that blocks this.

zopieux:
Eh, this CL celebrated its 4 year anniversary eleven days ago, so sadly I won't be giving it any more love. Anyone is very much welcome to take it over though.

Issue #158, opened 2017-02-07
void linux support


AMDmi3:
http://www.voidlinux.eu/packages/
https://github.com/voidlinux/void-packages

Another kind of shell scripts. Need preparsed metadata.

AMDmi3:
Alternatively, https://repo.voidlinux.eu/current/x86_64-repodata

AMDmi3:
Superseded by #400

PR #400, opened 2017-11-28
Add Void Linux repository support

ninewise:
I could use some feedback on `repos.d/voidlinux.yaml`, mainly the sources. I used a lot of subrepos because each source might have a different version, but perhaps I should use flavours for this? I'm not sure what's the accepted practice.

cc @Vaelatern


AMDmi3:
This data is not suitable for repology. There are incompatible versions with -N suffix (revision?), and there are subpackages (-32bit, -devel-32bit, -doc, etc.) which cannot be compared to anything. Having a lot of similar packages for all architectures is also a problem. See also #158.

NoctuaNivalis:
The revisions can easily be scratched (I believe `Package.origversion` could be used to refer to them?) and it shouldn't be hard to remove subpackages (provided I can get access to both a clone of the repository and the repodata file in the same parser?).

Should I then move the architectures to a flavour, would it still be possible to include voidlinux?

AMDmi3:
> The revisions can easily be scratched (I believe Package.origversion could be used to refer to them?)

Yes. See `repology/parsers/freebsd.py` for `SanitizeVersion` pattern used for this purpose. Please stick to it, as it would make refactoring easier in future.

> and it shouldn't be hard to remove subpackages (provided I can get access to both a clone of the repository and the repodata file in the same parser?).

You can't, and I don't understand how it's going to help you. However, as I see there's `source-revisions` field which may be useful: you could split it by `:` and set `package.effname` to the first part right away. That way, Void packages will be born assigned to correct metapackage, and information on package name will still be preserved.

> Should I then move the architectures to a flavour

No, the more correct would be to add flavor based on repository name. I've just added code to allow that.

> would it still be possible to include voidlinux?

As long as the data is of sufficient quality, but that can only be seen after proper parsing is implemented.

repology/parsers/voidlinux.py
```diff
@@ -0,0 +1,46 @@
+# Copyright (C) 2016-2017 Dmitry Marakasov <amdmi3@amdmi3.ru>
```
AMDmi3: Please set proper copyright with your name,

repology/parsers/voidlinux.py
```diff
@@ -0,0 +1,46 @@
...
+# You should have received a copy of the GNU General Public License
+# along with repology.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import plistlib
+
+from repology.package import Package
+
+
+def parse_maintainer(maintainerstr):
```
AMDmi3: There's already `from repology.util import GetMaintainers` tool.

repology/parsers/voidlinux.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __init__(self):
+        pass
+
+    def Parse(self, path):
+        index_path = os.path.join(path, 'index.plist')
+        plist_index = plistlib.load(open(index_path, 'rb'),
+                                    fmt=plistlib.FMT_XML)
+
+        return [Package(name=pkgname,
```
AMDmi3: Please fill fields one by one. The plans are to hide Package fields behind setters to catch incorrect data on assignment.

repology/parsers/voidlinux.py
```diff
@@ -0,0 +1,46 @@
...
+                        version=props['pkgver'].split('-')[-1].replace('_', '-'),
```
AMDmi3: Not sure if that's safe - `-` may be encountered in a version. I'd ensure that it starts with package name and strip `pkgname-` from the `pkgver`. Don't forget to strip revision.

repos.d/voidlinux.yaml
```diff
@@ -0,0 +1,22 @@
+###########################################################################
+# Voidlinux
+###########################################################################
+- name: voidlinux
```
AMDmi3: Is `linux' suffix necessary everywhere? I'd prefer to drop it for simplicity.

repos.d/voidlinux.yaml
```diff
@@ -0,0 +1,22 @@
...
+      subrepo: '{source}'
+  repolinks:
+    - desc: Void Linux home
+      url: https://www.voidlinux.eu/
+    - desc: Void Linux package repository
+      url: https://github.com/voidlinux/void-packages
+  packagelinks:
+    - desc: Git
+      url: 'https://github.com/voidlinux/void-packages/tree/master/srcpkgs/{name}'
+  tags: [ all, production ]
```
AMDmi3: Please ass newline

repos.d/voidlinux.yaml
```diff
@@ -0,0 +1,22 @@
...
+      parser: VoidLinux
+      url: https://repo.voidlinux.eu/current/{source}-repodata
+      subrepo: '{source}'
+  repolinks:
+    - desc: Void Linux home
+      url: https://www.voidlinux.eu/
+    - desc: Void Linux package repository
+      url: https://github.com/voidlinux/void-packages
+  packagelinks:
+    - desc: Git
```
AMDmi3: May also add another link to `template`

repos.d/voidlinux.yaml
```diff
@@ -0,0 +1,22 @@
...
+- name: voidlinux
+  type: repository
+  desc: Void Linux
+  family: voidlinux
+  sources:
+    - name: [ x86_64, armv6l, armv7l, i686, nonfree/armv6l, nonfree/armv7l, nonfree/i686, nonfree/x86_64, multilib/x86_64, multilib/nonfree/x86_64, musl/armv6l-musl, musl/armv7l-musl, musl/x86_64-musl, musl/nonfree/armv6l-musl, musl/nonfree/armv7l-musl, musl/nonfree/x86_64-musl, aarch64/aarch64-musl, aarch64/aarch64, aarch64/nonfree/aarch64-musl, aarch64/nonfree/aarch64, aarch64/nonfree/x86_64 ]
+      fetcher: WgetTar
+      parser: VoidLinux
+      url: https://repo.voidlinux.eu/current/{source}-repodata
+      subrepo: '{source}'
```
AMDmi3: `+      flavor: '{source}'` as mentioned in the comment

AMDmi3:
I still don't like these lots of repos. This produces excessive and duplicate package information which both slows down processing and makes website views cumbersome. Since you, as I assume, are Void Linux user, can't you set up native tools to produce machine readable index of source packages out of `void-packges` repository? That would make parsing easier, as unmangled package name and version are available and avoid duplicate packages.

AMDmi3:
I've meant preparing information on Void source packages _outside_ repology. Probably somewhere on Void Linux infrastructure.

Vaelatern:
The void-packages repository is simple: you list main packages by listing regular directories (not symlinks) and adding the python3 symlinks.

Packages do not exist on user systems in any way not merely from the repodata (including no subpkg vs. pkg distinction).

repology/parsers/voidlinux.py
Vaelatern: `-` may never be in a version.

NoctuaNivalis:
Latest version combines the void-package repository and the repodata. This way we can be sure which are actual packages and which are subpackages, while at the same time avoiding heuristics for parsing the template files.

For this to work, I had to write my own fetcher, and I supplied some extra variables to the `sources` to avoid hardcoding urls and flavors. This version creates a lot less packages compared to the previous, as it uses flavors for the architectures.

AMDmi3:
Looks overcomplicated now. Why's can't `source-revisions` be used?

AMDmi3:
I've tried it. The data looks mostly good, however for some reason `source-revisions` is missing for a fraction of packages. Is there a reason for it?

AMDmi3:
I've also noticed some subpackages with broken comment (seem to be replaced with suffix which was intended to be appended):

```
Nov 30 17:13:58 voidlinux:     sanity warning: libgadu-devel: comment is not stripped: " - development files"
Nov 30 17:13:59 voidlinux:     sanity warning: libsoxr-devel: comment is not stripped: " - development files"
Nov 30 17:13:59 voidlinux:     sanity warning: libsoxr-doc: comment is not stripped: " - documentation and examples"
Nov 30 17:14:00 voidlinux:     sanity warning: luaposix51: comment is not stripped: " - Lua 5.1"
Nov 30 17:14:00 voidlinux:     sanity warning: luaposix52: comment is not stripped: " - Lua 5.2"
```

Could it be that some packages need metadata reparsing?

AMDmi3:
See my branch: https://github.com/repology/repology/tree/voidlinux
The only problem left with it are packages without `source-revisions`. If this can be fixed from Void side, I'm merging it. Otherwise, we may investigate parsing raw `template` files with bashlex module I've just stumbled upon.

AMDmi3:
@ebfe from #voidlinux has a proof of concept json dump (https://gist.github.com/ebfe/e6ff186811cd207b884a9d7cff9d4e8a), I'd prefer to use it if it can be set up for automated regular updates anytime soon.

Chocimier:
Another option: output from `xbps-src` included in git repository output is quite readable:
```
~/github/void-packages ./xbps-src show netpbm
pkgname:        netpbm
version:        10.81.0
revision:       1
distfiles:      https://github.com/chneukirchen/netpbm-mirror/archive/a5fb397b3ab3ac6b2c17c72598036bedbdca1816.tar.gz
distfiles:      https://github.com/chneukirchen/netpbm-mirror/archive/ad3acab4d1ef5acd4a232c2bec3590620cd0edea.tar.gz
checksum:       ba1a03998d920f33a42450f583cd9024c90bb986c8b6a4ab7244d660deef593c
checksum:       55678f15a2ade5ed4c75f988a05296c895042e58a8c24829929562b3344a5349
maintainer:     Leah Neukirchen <leah@vuxu.org>
Upstream URL:   http://netpbm.sourceforge.net/
License(s):     BSD,GPL-2,custom
short_desc:     Toolkit for manipulation of graphic images
subpackages:    libnetpbm-devel
subpackages:    libnetpbm
```
Package is a subpackage if it lists itself in `subpackages` field.

AMDmi3:
@ebfe ping, any news? As a last resort, we can run json generation on Travis. Is the code which you used to generate https://gist.github.com/ebfe/e6ff186811cd207b884a9d7cff9d4e8a available somewhere?

Chocimier:
I found that no new build lacks `source-revision`, ie. last build without that record is from 2017-02-26.
There are just 109 packages that lacks the record in all architectures, so we can use [rules](https://gist.github.com/Chocimier/421319ae659fd2d62e94ba2ef663b26e#file-36-repo-void-yaml) for them, and try to share `source-revision` among architectures for the rest.

ebfe:
Please take a look at: https://gist.github.com/ebfe/71aacb55d7b17c8a72a8958fb26e8622

If this is enough information, i can look into hosting this somewhere.

ebfe:
> I found that no new build lacks source-revision, ie. last build without that record is from 2017-02-26.
> There are just 109 packages that lacks the record in all architectures, so we can use rules for them, and > try to share source-revision among architectures for the rest.

@Chocimier By my count there are still 5215 binary packages without the source-revisions property (3700 if you discount debug packages) from ~870 different templates.

XVilka:
Are there any updates on this one?

AMDmi3:
Void people seem to have lost interest. I'll review what we have when I return from vacation in 2 weeks. Also maybe we can set up Travis job to bake us more useful data.

ebfe:
> Void people seem to have lost interest.

If someone had responded to
https://github.com/repology/repology/pull/400#issuecomment-370145650 it
would already be there. These days i'm not spending a lot of time on
void so you have to find someone else.

AMDmi3:
Hm, either I've somehow managed to miss it, or it wad discussed it on IRC. Well it woild be nice if it was valid JSON (array of dicts, not concatenation of dict JSONs). And if it's intended solely for repology, revision and binpkgs would be useless and could be removed. Also if it's possible to include source download locations and some kind of package category, that would be useful too.

Well the only show stopper is actually JSON incompatible format.

AMDmi3:
@ebfe any progress?

ebfe:
As written above, i don't work on this anymore.

AMDmi3:
Oh, ok.

Chocimier:
@ebfe, could you share your tool generating json?

the-maldridge:
A few things to add on here.  bashlex isn't really a good option here because templates aren't guaranteed to be parsable with static analysis methods (this is something we're actively evaluating fixes for).

Void doesn't have package categories and so that is a field that will be missing.

Chocimier:
We are rebuilding packages to generate missing `source-revisions`, more than half is done. Note that for some packages there is trailing `/template` there.

We moved website to https://voidlinux.org , github to https://github.com/void-linux/void-packages and repository to https://alpha.de.repo.voidlinux.org/current/ .

AMDmi3:
Lets recap this. The code is a year old rotten and needs to be rewritten from scratch. I no longer remember details so I have no idea what's `source-revisions` is about. I can't even examine current repository contents as it acts like it knows better than me whether I can handle a file list. The latest version of code wanted to parse 21 repodatas which would presumably lead to a lot of duplicated packages, if that's still the case that is not acceptable.

Chocimier:
> I no longer remember details so I have no idea what source-revisions thing is about.

Just useful metadata field missing then ([details](https://github.com/repology/repology/pull/400#issuecomment-347894366)).

> I can't even examine current repository contents

Bothers me too, but parser is based on repodata files, not directory listing.

> The latest version of code wanted to parse 21 repodatas which would presumably lead to a lot of duplicated packages, if that's still the case that is not acceptable.

We fixed repo as, you [were](https://github.com/repology/repology/pull/400#issuecomment-348219657) about to accept it.

> a lot of duplicated packages

If it is a problem, I am sure there is a way to handle that: keeping entry non-duplicated by database, having a set of names of already-registed packages in parser or something other.

> I also assume there's still no dedicated json

You are right, there is no.

AMDmi3:
> Bothers me too, but parser is based on repodata files, not directory listing.

I cannot maintain it unless I see the repository contents at least.

> We fixed repo as, you were about to accept it.

As the time pass, and more repositories are supported, quality requirements raise, as each problematic repository creates more work for me and more noise for the users.

> If it is a problem, I am sure there is a way to handle that: keeping entry non-duplicated by database, having a set of names of already-registed packages in parser or something other.

Not an option, as this introduces inconsistencies.

However, we could stick with e.g. `x86_64` + `nonfree/x86_64`.

AMDmi3:
> Bothers me too, but parser is based on repodata files, not directory listing.

I cannot maintain it unless I see the repository contents at least.

> We fixed repo as, you were about to accept it.

As the time pass, and more repositories are supported, quality requirements raise, as each problematic repository creates more work for me and more noise for the users.

> If it is a problem, I am sure there is a way to handle that: keeping entry non-duplicated by database, having a set of names of already-registed packages in parser or something other.

Not an option, as this introduces inconsistencies.

However, we could stick with e.g. `x86_64` + `nonfree/x86_64`.

AMDmi3:
> Bothers me too, but parser is based on repodata files, not directory listing.

I cannot maintain it unless I see the repository contents at least.

> We fixed repo as, you were about to accept it.

As the time pass, and more repositories are supported, quality requirements raise, as each problematic repository creates more work for me and more noise for the users.

> If it is a problem, I am sure there is a way to handle that: keeping entry non-duplicated by database, having a set of names of already-registed packages in parser or something other.

Not an option, as this introduces inconsistencies.

However, we could stick with e.g. `x86_64` + `nonfree/x86_64`.

AMDmi3:
> Bothers me too, but parser is based on repodata files, not directory listing.

I cannot maintain it unless I see the repository contents at least. Transparency is an absolute requirement.

> We fixed repo as, you were about to accept it.

As the time pass, and more repositories are supported, quality requirements raise, as each problematic repository creates more work for me and more noise for the users.

> If it is a problem, I am sure there is a way to handle that: keeping entry non-duplicated by database, having a set of names of already-registed packages in parser or something other.

Not an option, as this introduces inconsistencies.

However, we could stick with e.g. `x86_64` + `nonfree/x86_64`.

Chocimier:
> However, we could stick with e.g. x86_64 + nonfree/x86_64.

Fine for me. Is listing directory only major issue (while I do not understand why is it an issue) then?

AMDmi3:
> Fine for me. Is listing directory only major issue (while I do not understand why is it an issue) then?

For now, yes. I still need to check the data.

the-maldridge:
Seems like there's some serious confusion here about how xbps works and how our mirror infrastructure works.

First, there are multiple repodata files because there are multiple repositories.  If you don't parse them then you won't have the complete view of the universe.  If you're okay having massive holes in the dataset, then just parse a handful of files.

Second, the current/ directory is for machines, not humans.  The page is there to prevent you from loading a page with 10k+ links on it which your browser will try to prefetch for validity.  If you believe you actually want a file list then get it from rsync which is designed to do that, or get yourself netblock banned from someone else's mirror.  There's no transparency in that list anyway, its pretty trivial to filter for specific UAs and give back a different list there.

Vaelatern:
If you need to check the data, you can compose a URL to any package in question yourself. This is straightforward, and the package is provided as a tarball (by a different extension, but a tarball).

the-maldridge:
> The completeness of the dataset should bother you, as a primary service consumer, in the first place.

I am not a consumer, I'm the person on the other end who's dealing with people hammering mirrors rather than parsing the machine readable data.

> I need to see what's I'm working with to keep it working and improve. Mirrors do not help as they all serve the stub instead of a listing. It would be OK if it worked with curl.

Then summon the listing via rsync as suggested.

Vaelatern:
> But, since there is no source packages (e.g. templates) metadata, and I'm not going to just dump a whole bunch of duplicates (archs + musl + multilib) for each package into repology, that's what we'll have to live with.

How would you prefer to get all the data from archs, musl, nonfree?

AMDmi3:
> Then summon the listing via rsync as suggested.

Example commandline please?

AMDmi3:
> How would you prefer to get all the data from archs, musl, nonfree? Since any format will have duplicates that don't share every quality (same pkgname version, different arch)

I'd prefer not to deal with binary package data at all.

the-maldridge:
```
rsync rsync://alpha.de.repo.voidlinux.org/voidmirror/current
```

I suggest piping that to less.

the-maldridge:
Well, if you don't deal with the binary package data then you'll get bogus data.  Not all packages sitting in the webroot are in the current generation of the tree.

AMDmi3:
> `rsync rsync://alpha.de.repo.voidlinux.org/voidmirror/current/`

Thanks, it needs a trailing slash, but it's acceptable for my purpose.

> Well, if you don't deal with the binary package data then you'll get bogus data. Not all packages sitting in the webroot are in the current generation of the tree.

It's not bogus, it's just the discrepancy between source and binary packages. Ideally, we need knowledge of both, so we can both tell maintainers which packages need work (tied to commits), and tell users which packages are already available for them (tied to package builds). Software authors are probably interested in both. However for now we focus on maintainers, who want outdated status of their packages to be cleared as soon as possible after they commit the update. The lag of actual package builds is expected and considered tolerable, no one complained on that for now.

the-maldridge:
As long as you understand that the tree probably doesn't resolve at all times, I guess that's fine.  I'm not sure how your system works then without referential integrity checks.

You should probably pick another mirror though to pull that rsync listing from, unless of course your machines are in Germany.

Vaelatern:
@AMDmi3 there are often multiple versions of a package in the physical directory, though only one in the repo index. The only thing that matters to users or maintainers is what's in the index. Everything else is irrelevant.

Vaelatern:
To be clear. The repo index has nothing to do with the theoretical packages we have. It has EVERYTHING to do with the packages that people can actually install at that very moment.

AMDmi3:
> The only thing that matters to users or maintainers is what's in the index.

That's definitely not true for maintainers as they commit to the repository and the state of repository is what is relevant to them.

Vaelatern:
@AMDmi3 As someone with close knowledge about Void operations, you should know that the only thing that matters is what's in the repo index, and that repo index is authoritative.

Chocimier:
> That's definitely not true for maintainers as they commit to the repository and the state of repository is what is relevant to them.

Nope. When template is updated and binary index is not updated for time longer than usual processing, it is clear signal that template is broken in some way causing new version of program not being available for users. Therefore it is not important what template contain. Binary index is.

AMDmi3:
@Chocimier JFYI, there are some minor parsing problems, including misformatted URLs and couple of missing source-revisions.

https://repology.org/log/550490

Chocimier:
Missing source-revisons come from packages troublesome to rebuild, as I mentioned before. Homepages are fixed now, thanks for info.

Issue #166, opened 2017-02-15
Implement simple shell parser


AMDmi3:
We need to parse shell constructs to extract variables from a lot of package formats (such as ebuilds, pkgbuilds etc). ~~May use shlex python module~~.

Repositories which will be allowed/improved:
- HaikuPorts (#260)
- Void linux (#158, #301)
- Gentoo
- MSYS2 (#262)
- OpenELEC (#205)
- FrugalWare (#169)
- SliTaZ (#1)
- NuTyX  (#1)
- KaOS (#323)
- Chakra linux (#327)
- Source Mage (#364)
- Termux (#406, #628)
- Exherbo (#630)

Symbian9:
SliTaz --> SliTaZ

Void --> Void Linux

AMDmi3:
This module turned out to be garbage not even being able to process shell files with comments.

mvdan:
This might be of use to you: https://github.com/mvdan/sh

It's not written in Python, but I believe it is possible to build a Go package as a C shared library and call it from most languages, including Python. For example: http://savorywatt.com/2015/09/18/calling-go-code-from-python-code/

The binary also has a json output flag, which might help bring the barrier down:

```
$ cat foo.sh
echo foo bar # comment
$ shfmt -exp.tojson <foo.sh
<AST as a JSON object>
```

AMDmi3:
For the note, this is unlikey to be implemented anytime soon. Even if it is, there's no guarantee that scripts are be parsed correctly, as they may involve any kind of complex logic and call any kind of unportable non-deterministic utilities, so I don't treat it as a viable solution at all.

So official verdict is that we won't support repositories unless they provide metadata in machine readable format.

mvdan:
@AMDmi3 if this is at all useful, the library also includes a shell interpreter, so you can safely source a shell file with it: https://godoc.org/mvdan.cc/sh/shell#SourceNode

If you look at the code, you'll see that it's whitelist-based - it only allows running a few well-known coreutils programs, and forbids access to disk. You could also customize that however you want.

AMDmi3:
The note was not related to your project, but general - I'd like to simplify repology logic and dependency tree, and using elaborate parsers for specific repos doesn't seem the right solution. Repos should provide machine readable metadata.

AMDmi3:
Scrap this.
- It requires too much effort
- It may still not solve the problem, as most shell-based build systems have huge infrastructure which need to be parsed and likely executed as well, and it depends on environment
[...]

PR #406, opened 2017-12-12
Add Termux repository

stephengroat:
no sources.gz available, wondering if a packages.gz will work


AMDmi3:
Unfortunately it won't, as it's polluted with subpackages and arch-dependent repositories are likely to miss some packages. Since custom build scripts are used we'll only be able to parse this with #166.

PR #407, opened 2017-12-12
create python standard requirements.txt

stephengroat:


AMDmi3:
This is not completely correct. At least flake8* and coveralls are not needed for normal run, so they are not needed in requirements.txt. Some other depends are optional too.

AMDmi3:
Closing due to no feedback.

PR #425, opened 2018-01-02
add `rindeal` ebuild repository

rindeal:


AMDmi3:
Sorry, but repology policy is not to add private repositories.

rindeal:
What do you mean by _private_?

AMDmi3:
These maintained by a single person.

rindeal:
Linuxunderground overlay is _private_ in your sense of the word yet it's there, but nevermind. Still maybe you could state the policy in some docs, so that others won't waste time like me.

AMDmi3:
Fair enough, it will be removed.

PR #499, opened 2018-02-14
Add more devel rules

jtojnar:
Adding `gnome-clocks`, `gnome-contacts`, `gnome-user-share` and `nautilus`.


jtojnar:
Added `orca`, `gnome-documents` and `baobab`.

PR #565, opened 2018-03-27
Problems page linking

### LINKED ISSUE #512: Add direct links to the packages in the "repository problems" page
The "Repository problems" page ([example](https://repology.org/repository/wikidata/problems)) includes the package name but not the link. For contrast, see the "Versions" page ([example](https://repology.org/metapackage/0ad/versions)), which does contain a direct link to the package's page in the repository's website. Having those links in the "Repository problems" page would be convenient for addressing the issues.

I tried to submit a PR but got a bit lost in the code. From:

https://github.com/repology/repology/blob/840c1ba6b0d7ddc67d4fd1070cd8f96aef1a60bd/repologyapp/templates/metapackage-versions.html#L34

...I can see that the `package_link()` macro is used to produce the link in the "Versions" page. However, in the "Repository problems" page [the object is a `problem`](https://github.com/repology/repology/blob/840c1ba6b0d7ddc67d4fd1070cd8f96aef1a60bd/repologyapp/templates/problems.html#L25), which doesn't sound like the kind of object `package_link()` would expect.

Granted, I noticed from:

https://github.com/repology/repology/blob/252fc436e665a46d084614dcad623a4a8c669b77/repologyapp/templates/macros.html#L29-L35

...that the first argument to the macro would need to have a `repo` field, which apparently the `problem` data struct [does](https://github.com/repology/repology/blob/8abe96826ee490f152f39919b7c8c559d99ab47a/sql.d/problems.sql#L35); however, it apparently would also need to be a data structure that can be passed to [`pkg_format`](https://github.com/repology/repology/blob/371fb0976d74267291f448ebcfa928820a884581/repologyapp/template_helpers.py#L75)... I didn't follow the trail further since it already feels a bit hackish to pass a `problem` object to `package_link()`, rather than a `package` object.

I'd welcome any hints as to how to proceed.

nemaniarjun:
Closes #512

CC: @jayvdb, @AMDmi3, @waldyrious


waldyrious:
Thanks for tackling this! I'm afraid I have neither the technical background nor the availability to give this a proper review, but let me know if I can help with testing or something else.

sql.d/schema.sql
```diff
@@ -408,6 +409,7 @@ CREATE TABLE links (
 -- Problems
 --------------------------------------------------------------------------------
 CREATE TABLE problems (
+    package_id int NOT NULL,
```
waldyrious: Minor nit: the indentation on this line seems to be inconsistent with the ones below (looks like a tabs vs. spaces issue)

repology/repoproc.py
```diff
@@ -309,7 +309,11 @@ def DeserializeMulti(self, reponames=None, logger=NoopLogger()):
     def StreamDeserializeMulti(self, processor, reponames=None, logger=NoopLogger()):
         deserializers = []
         for repo in self.repoman.GetRepositories(reponames):
-            deserializers.append(self.StreamDeserializer(self.__GetSerializedPath(repo)))
```
AMDmi3: Please remove unrelated changes: this

repology/subprocess.py
```diff
@@ -31,8 +31,6 @@ def RunSubprocess(command, logger, cwd=None):
                           stdout=subprocess.PIPE,
                           stderr=subprocess.STDOUT,
                           universal_newlines=True,
-                          encoding='utf-8',
```
AMDmi3: and this

sql.d/schema.sql
```diff
@@ -369,7 +370,7 @@ CREATE TABLE repositories (
 -- repository_history
 CREATE TABLE repositories_history (
 	ts timestamp with time zone NOT NULL PRIMARY KEY,
-	snapshot jsonb NOT NULL
```
AMDmi3: and this

sql.d/schema.sql
```diff
@@ -146,6 +146,7 @@ $$ LANGUAGE plpgsql;
 -- Main packages table
 --------------------------------------------------------------------------------
 CREATE TABLE packages (
+	id serial,
```
AMDmi3: Please use postgresql 10 new way of specifying identity columns, as here:

https://github.com/repology/repology/blob/694db62a67e517bef75e5ec5881973fbc4a394d0/sql.d/schema.sql#L428

sql.d/schema.sql
nemaniarjun: Thanks! Will fix :)

repology/repoproc.py
nemaniarjun: Yep, knew these were unrelated thus put them in a seperate commit. :)
Couldnt have tested the solution and these changes were what stopped errors for me.

nemaniarjun:
Updated as per comments. The Travis build is failing though.

AMDmi3:
It fails because `Package` class constructor is fed with a new `id` field it doesn't know about. You may fix this by specifying field names to select in `sql.d/packages.sql` explicitly (instead of `*`). You can get the list of fields from INSERT statement in the very same file.

AMDmi3:
Great, looks fine. Will merge after some testing.

sql.d/problems.sql
```diff
@@ -31,14 +31,13 @@ WHERE repo = %(repo)s;
...
-	name,
-	effname,
+SELECT 
+	packages.*,
 	maintainer,
 	problem
 FROM problems
-WHERE repo = %(repo)s
+INNER JOIN packages ON packages.id = problems.package_id
+WHERE packages.repo = %(repo)s
```
AMDmi3: Post-mortem note: this was not correct - you've had to use problems.repo here. Despite that the contents are the same, if you refer to `packages` for conditions or sorting, no the index scan on `problems` cannot be used and the database fallbacks to seq scan which leads to problems page answering much longer. Also you've missed the update to similar `get_maintainer_problems` query.

PR #599, opened 2018-04-29
update entware repository link in news

davidak:
Not sure if rewriting history is ok.

At least the link should be fixed. Currently 404!

Related to https://github.com/repology/repology/pull/554.

### DIFF
--- a/repologyapp/templates/news.html
+++ b/repologyapp/templates/news.html
@@ -37,7 +37,7 @@
 <article>
 	<h3 id="2018-02-16">2018-02-16</h3>
 	<ul>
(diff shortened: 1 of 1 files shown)


AMDmi3:
From one hand, the link is dead because repository was renamed. From another, the repository _was_ entware-ng at the time of writing. Let's leave the link text as is, it seems to be the most correct solution.

davidak:
That's a good solution. I changed it. Sadly i was not able to `--amend` in github webinterface.

AMDmi3:
:+1:

> Sadly i was not able to --amend in github webinterface.

It's ok, since there's squash merge option. Never used it before :)

AMDmi3:
FYI, there are holidays currently in Russia, so I'll review your issues in some days.

Issue #527, opened 2018-03-05
Track package lifetimes through keynames


AMDmi3:
It's useful to track temporal data no packages, however it's not possible with either `packages`, as there's no way to track a single package between repository updates, neither `metapackages` and these are prone to fluctuations caused by rule changes.

We can introduce an intermediate entity.
- [ ] first, introduce `keynames` field to all packages (#439) which is an alternative name which is not prone to ambiguities and is stable across updates. For example, for Gentoo it would be a path in the portage tree (e.g. `dev-haskell/async`, `dev-ml/async`, `dev-python/async` as opposed to `async`).
- [ ] add a table with (`repo`, `keynames`) primary key which will track `first_seen`, `last_seen` fields and possibly other historical data instead of `metapackages` table. We need a way to tie that with `effname`s to make it usable though.
- [ ] use it in actual queries

AMDmi3:
Done in #1000, experimental application is https://repology.org/experimental/trending

PR #616, opened 2018-05-06
Add a redirect from repo package names to metapackages

pabs3:
This will be useful for distros trying to link to metapackages.

### DIFF
--- a/sql.d/get_package_metapackage.sql
+++ b/sql.d/get_package_metapackage.sql
@@ -1,0 +1,30 @@
+-- Copyright (C) 2018 Paul Wise <pabs3@bonedaddy.net>
+--
+-- This file is part of repology
+--
+-- repology is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation, either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- repology is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


sql.d/get_package_metapackage.sql
```diff
@@ -0,0 +1,30 @@
...
+
+--------------------------------------------------------------------------------
+--
+-- @param repo
+-- @param package
+--
+-- @returns array of metapackages
+--
+--------------------------------------------------------------------------------
+SELECT
```
AMDmi3: This query would be slow unless suitable index is also added.

AMDmi3:
I've got the point of these PRs (#615 as well), but I can't merge them right away. First, they break the code (see travis). Next, we need to think of consistent URL scheme here. IMO both should be under `/repository` as package names are repository specific, e.g. `/repository/<repo>/package/<package>/*` where is is either
- Everything supported by `/metapackage/*` subtree (redirects)
- `problems` (new page)

Next, `flask.abort(500)` in case of multiple packages is not acceptable. It'll fire when there are multiple versions of the same package which is pretty common case, so first you need to count distinct `effname`s instead of metapackages, and provide disambiguation page if there are multiple of them.

Also, I've planned to implement _origins_ (#527), which are alternative unique package names (may be the same as names for some repos, but totally different for others, for instance on FreeBSD there are packages named `py27-Flask` and `py36-Flask` with origin `www/py-flask`), I think it would be more suitable here.

pabs3:
My SQL is a little rusty so I need some help modifying the
query_metapackages SQL to support searching the package names.

-- 
bye,
pabs

http://bonedaddy.net/pabs3/

PR #626, opened 2018-05-26
add categories to nix parser

ryantm:
related: https://github.com/NixOS/nixpkgs/issues/41029

### DIFF
--- a/repology/parsers/nix.py
+++ b/repology/parsers/nix.py
@@ -89,6 +89,10 @@ class NixJsonParser():
                         pkg.name = pkgname
                         pkg.version = merged[len(pkgname) + 1:]
 
+                keyparts = key.split('.')
+                if len(keyparts) > 1:
+                    pkg.category = keyparts[0]
+
                 if pkg.name.endswith('-git'):
                     pkg.name = pkg.name[:-4]
                     pkg.SetFlag(PackageFlags.ignore)


ryantm:
@AMDmi3 Okay, I think this will work! nixpkgs doesn't have categories, so I'm setting the category based on the sub package set. For example, if the key is `haskellPackages.errors`, the category will be set to `haskell`. This also means some packages will have categories like "python36", but that seems okay to me. After this goes live, we can add an entry to `500.wildcard.yaml` to fix the `haskellPackages`.

AMDmi3:
Hmm. I've planned to fix this by adding `haskell:` prefix if the key starts with `haskellPackages.` but name doesn't start with `haskell-`. One one hand, your solution is more generic and exposes information into rules, which is good, but I'd prefer to avoid a lot of package names jumping back and forth if/when `haskell-` prefix is added. Maybe we should fix names my way, but merge this anyway to enrich package information with category, what do you think? Btw, isn't there more general pattern in these keys? E.g. can't we take everything up to the first dot? E.g. `haskellPackages` as a cstegory, and I've seen `vim` and other ones.

ryantm:
@AMDmi3 I think taking everything up to the `.` is a good idea! I looked through everything that would include and it seems fine, and I updated this PR to do that.

Are you proposing adding your other fix to the parser too? I'm fine with that, but it seems like we can cover it with a rule if the first part of the key is in the category.

AMDmi3:
Yes, but if/when prefix is added we'll have names like `haskell:haskell-error` where first one is added by the category, and the second is name prefix added in Nix. On the second thought, it can be fixed with rules as well (with a regexp lookahead or with 3 rules which replace prefix (`haskell-` -> `haskell:`, then add `haskell:` based on category, then remove duplicate prefix).

Btw, there already are a handful of packages prefixed by `haskell-`, could you please investigate them? Is it ok to process the prefix right away, or should thesr be added to exceptions?

ryantm:
I think we should process the prefix right away, there are thousands of metapackages that are currently messed up by this. Sure, I'll look into the ones that already have the prefix.

ryantm:
It looks like most or all of those packages prefixed with "haskell-" are actually named that on Hackage. For example, https://hackage.haskell.org/package/haskell-abci.

AMDmi3:
This complicates things. Then maybe we could wait for Nix community to decide on whether haskell packages are going to be prefixed or not. I'd really prefer to avoid massive name changes, this pollutes feeds.

ryantm:
How does it complicated things? If nixpkgs adds the prefix, that example should be `haskell-haskell-abci`, and if Repology strips off the first `haskell-` that's no big deal.

ryantm:
Oh, were you saying you would preemptively add the prefix stripping rule?

AMDmi3:
Yes

ryantm:
There is debate about whether to prefix the Haskell packages, and I'm not sure it will be resolved quickly. Meanwhile, the nix Haskell packages are mixing in with other packages where they don't belong. For example: 

* https://repology.org/metapackage/mtl/versions
* https://repology.org/metapackage/pipes/versions

I found those examples in only a couple minutes of searching, so I bet there are easily many more.

I think fixing wrong information in the system is more important than keeping the feeds clean, so I think we should merge this and add the rule that the category `haskellPackages` in the `nix` family should be prefixed by `haskell:`.

AMDmi3:
Well OK, if it's not going to be decided right away, let's use categories, as false package aliasing is really critical problem. However, let's simplify it a little bit - I think `split()` is enough here. As in `parts = key.split('.')`, `if len(parts) > 1: pkg.category = parts[0]`

ryantm:
@AMDmi3 done!

PR #646, opened 2018-06-21
exherbo.yaml: add dev repositories

Cogitri:


AMDmi3:
These should definitely not be mixed with official Exherbo repositories. Also, Repology policy is strongly against private repositories - these are not directly accessible to users, are underreviwed, prone to contain outdated, low quality and broken packages, and tend to become abandoned.

Cogitri:
> These should definitely not be mixed with official Exherbo repositories
Alright, sure

> Also, Repology policy is strongly against private repositories - these are not accessible to users out of box, are underreviwed, prone to contain outdated, low quality and broken packages, and tend to become abandoned.

Hm, this doesn't mix well with Exherbo's way of dealing with repositories. We actually encourage mixing official, dev and unofficial repos because we don't duplicate any packages between them. Unofficial repositories are basically an easy way to introduce new packages which aren't important enough to be in one of the official ones. Unofficial repositories get reviewed by the core devs and are abandoned if they don't meet our quality standards. Unofficial/Dev repos are as accessible as any other kind of repository which is available for Exherbo (you can install it with our package manager - cave resolve repository/<package>). At least dev repositories receive the same amount of work and care as offical repositories (since they're both maintained by the same people - all exherbo users)

repos.d/exherbo.yaml
```diff
@@ -46,6 +46,58 @@
...
+        - moben
+        - mrothe
+        - olesalscheider
+        - philantrop
+        - pioto
+        - pipping
+        - pyro
+        - rbrown
+        - replica
+        - rejo
```
AMDmi3: Shouldn't it be `sejo`?

AMDmi3:
> We actually encourage mixing official, dev and unofficial repos because we don't duplicate any packages between them

That's unusual, but that's whole another story. In this case the dev repos set is good. Would still prefer unofficial ones to be removed.

AMDmi3:
Oh, forgot one more thing: let's just add them to existing section as dev/*.

Cogitri:
Sure, I feel like it looks a little cleaner otherwise, but if you say so

Cogitri:
Argh, typo

AMDmi3:
> Sure, I feel like it looks a little cleaner otherwise

Well, this way there's less duplication and `subrepo` is filled with more informative `dev/<repo>` instead of just `<repo>`.

AMDmi3:
FYI, extended set of repositories is now in production.

Cogitri:
Okay, thanks :)

PR #685, opened 2018-09-01
Add Adélie Linux

sroracle:


AMDmi3:
Could you please share some details on the distro? How does it relate to alpine?

sroracle:
We were previously a pseudo-fork of Alpine but we no longer track them as an upstream anymore. We still use their package manager and build system, however. In terms of goals, Adélie seeks to become POSIX certified and to serve architectures that have been relatively unmaintained in other distributions (such as ppc32be, ppc64be, x86_32, etc.) in addition to the ubiquitous x86_64. We are also planning on supporting sparc64 and mips in the future. I encourage you to check out [our homepage](https://adelielinux.org) if any of these things interest you.

AMDmi3:
Well, looks good to me. But why ppc64 arch? If I were to pick one I'd pick x86_64 as probably it has the most packages available. Can also use multiple archs, but that'd be a bit clumsy and probably won't bring much benefit. x64_64 is used for Alpine.

Also, are there any package sources (e.g. build recipes or something lile that) available? Would be nice to at least link to them.

Finally, how far is it currently from alpine? Rough estimation by numbers of matching versions and quick glance on some packages gave me an idea there are little on no similar packages now,, which means it'd be better to use individual `family` and `ruleset` values for Adélie. Still need to check whether Alpine-specific rules apply here though.

repos.d/adelie.yaml
```diff
@@ -0,0 +1,20 @@
+###########################################################################
+# Adélie Linux aports
```
awilfox: Packages, not aports

repos.d/adelie.yaml
```diff
@@ -0,0 +1,20 @@
...
+  sources:
+    - name: [ system, user ]
+      fetcher: WgetTarFetcher
+      parser: ApkIndexParser
+      url: 'https://distfiles.adelielinux.org/adelie/current/{source}/ppc64/APKINDEX.tar.gz'
+      subrepo: '{source}'
+  repolinks:
+    - desc: Adélie Linux home
+      url: https://adelielinux.org/
+  tags: [ all, production ]
```
awilfox: We're still in beta.  is 'production' the correct tag?

awilfox:
The ppc64 arch has the most packages available, and is our primary architecture.

[[sroracle]]'s explanation of tier 1 is correct, but I'd like to note that specifically there are packages that are failing tests and therefore not built on x86_64 and pmmx (x86_32) where they pass and are built on ppc64.  (We have until the beta1 window closes to fix them.)  That's why I said ppc64 should be the URL for use.

repos.d/adelie.yaml
sroracle: I took this to mean that it is available on repology.org. @AMDmi3 thoughts?

AMDmi3: This of course relates to the status of repository in Repology - `production` denotes that the repository is available on [repology.org](https://repology.org) instance. We don't care of upstream production status.

AMDmi3:
Thanks for ppc64 related clarification and the link to the recipes!

Regarding `family` and `ruleset`. These are used to group related or similar repositories together. `family` is mainly used for statistics, including the calculation of package `spread`. E.g. we want a metric of how many repos a package is available in, but we don't want to count it multiple times for e.g. all Debian, Ubuntu and other Debian-derived distros, because all these basically share the same package. Similarly, `ruleset` is used to limit rules by specific repos, but likewise we don't want to mention all of them literally, so we refer to them via `ruleset` group name.

I've checked and it looks like Adèlie and Alpine no longer have much common packages, so Adèlie needs dedicated `family` and `ruleset`. Please change them both to `adelie` and we're good to go.

sroracle:
Okay, I changed `family` and `ruleset` to `adelie` and added `adelie` to `tags`.

PR #738, opened 2018-10-30
Add Fedora 29

c72578:


AMDmi3:
It replies with 403

https://mirror.yandex.ru/fedora/linux/releases/29/

c72578:
Fedora 29 was officially released on 2018-10-30, so it is only a question of time, until mirror.yandex.ru is synchronized and permission granted.

c72578:
It is OK now.

AMDmi3:
Merged

PR #847, opened 2019-05-12
Update Termux metadata url

fornwall:
Termux is switching to use bintray for package hosting.

The old repository at termux.net will remain in maintenance mode for now, but active development happens at bintray.

### DIFF
--- a/repos.d/termux.yaml
+++ b/repos.d/termux.yaml
@@ -11,7 +11,7 @@
     - name: packages.json
       fetcher: FileFetcher
       parser: TermuxJsonParser
-      url: https://termux.net/dists/stable/packages.json
+      url: https://dl.bintray.com/termux/metadata/repology/packages.json
   repolinks:
     - desc: Termux home
       url: https://termux.com/


fornwall:
By the way, thanks for the awesome work you do in repology, it's a great service which helps keeping packages updated!

hazemyasr:
hey fornwall i downloaded termux last night and i have a problem when i try to installing "pip install numpy" it's gives me a big error and i try "apt update && apt upgrade" and it didn't work

hazemyasr:
> Termux is switching to use bintray for package hosting.
> 
> The old repository at termux.net will remain in maintenance mode for now, but active development happens at bintray.

hey fornwall i downloaded termux last night and i have a problem when i try to installing "pip install numpy" it's gives me a big error and i try "apt update && apt upgrade" and it didn't work

AMDmi3:
@fornwall  Hey, I've noticed that we now have some broken links to package sources, such as jp2a link leading to https://github.com/termux/termux-packages/tree/master/packages/jp2a. As it turns out, there's separate repository now at https://github.com/termux/packages-unstable, and it should point there, but there's no way to know that from the data we use.

Can source repository info be added to the dump, or can it be split into per-repository dumps? The latter would also be useful to supply accurate information on packages available in stock Termux vs. packages available in repositories not enabled by default (e.g. unstable ones). As I can see in my Termux, `jp2a` is not returned by `pkg search jp2a`

PR #857, opened 2019-05-27
Add support for Cygwin repo

### LINKED ISSUE #269: Cygwin support
@tkelman:
> Cygwin: http://mirrors.mit.edu/cygwin/x86_64/setup.ini, http://mirrors.mit.edu/cygwin/x86/setup.ini

jon-turney:
Closes #269

### DIFF
--- a/repology/parsers/parsers/cygwin.py
+++ b/repology/parsers/parsers/cygwin.py
@@ -1,0 +1,66 @@
+# Copyright (C) 2019 Jon Turney <jon.turney@dronecode.org.uk>
+#
+# This file is part of repology
+#
+# repology is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# repology is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with repology.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Generator
+
+from jsonslicer import JsonSlicer
+
+from repology.package import PackageFlags
+from repology.packagemaker import PackageFactory, PackageMaker
+from repology.parsers import Parser
+from repology.transformer import PackageTransformer
+
+
+class CygwinParser(Parser):
+    def iter_parse(self, path: str, factory: PackageFactory, transformer: PackageTransformer) -> Generator[PackageMaker, None, None]:
+        with open(path, 'rb') as jsonfile:
+            for packagedata in JsonSlicer(jsonfile, ('packages', None)):
+                # packages with names starting with an underscore are
+                # uninteresting as they contain cygwin-specific
+                # installation helper scripts
+                if packagedata['name'].startswith('_'):
+                    continue
+
+                pkg = factory.begin()
+
+                pkg.set_basename(packagedata['name'])
+                pkg.set_summary(packagedata['summary'])
+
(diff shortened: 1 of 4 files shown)


repology/parsers/parsers/cygwin.py
```diff
@@ -0,0 +1,64 @@
...
+                    continue
+
+                pkg = factory.begin()
+
+                pkg.set_basename(package['name'])
+                pkg.add_categories([c.lower() for c in package['categories']])
+                pkg.set_summary(package['summary'])
+
+                if 'maintainers' in package:
+                    pkg.add_maintainers([m.lower().replace(' ', '.') + '@cygwin' for m in package['maintainers']])
```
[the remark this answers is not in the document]
jon-turney: Well spotted

jon-turney:
> Some minor changes suggested,

Updated to incorporate those suggested changes

> also a question: I see `.cygport` files contain `HOMEPAGE` and `SRC_URI`, couldn't these be propagated to the dump? 

Unfortunately there's an intermediate step which doesn't propagate those pieces of information, currently.  

> Also isn't there a way to have/generate a link to source `.cygport` for each package?

Unfortunately, for historical reasons, there's no canonical repository for our cygport files, currently.

Though both things would be good to add, certainly :)

repology/parsers/parsers/cygwin.py
```diff
@@ -0,0 +1,63 @@
...
+                for maturity in ['stable', 'test']:
+                    if maturity not in packagedata['versions']:
+                        continue
+
+                    version = packagedata['versions'][maturity][-1]
+
+                    for subpackage in packagedata['subpackages']:
+                        pkg = factory.begin()
+
+                        pkg.set_basename(packagedata['name'])
```
AMDmi3: Sorry, perhaps I wasn't too clear. I haven't meant generic fields like basename, categories, summary and maintainers to be set in the inner loop. They belong outside of the loop, and this makes logs produced by PackageMakager more useful as basename can be used to identify the package which produces output. In the inner loop, just `clone` pkg and set subpackage specific fields like version, flags and name.

repos.d/cygwin.yaml
```diff
@@ -0,0 +1,24 @@
+###########################################################################
+# Cygwin
+###########################################################################
+- name: cygwin
+  type: repository
+  desc: Cygwin
+  family: cygwin
+  color: '00fb15'
+  minpackages: 7000
```
AMDmi3: There are packages with no maintainers defined (at least `alternatives`). You may want to set e.g. `default_maintainer: orphaned@cygwin` here to be able to list these in repology.

repology/parsers/parsers/cygwin.py
jon-turney: Ah, right.  I think I understand what you meant now.

jon-turney:
Moved the check for '_obsolete' category from package to subpackage.  Among the effects this was having was suppressing a lot of python packages.

repology/parsers/parsers/cygwin.py
```diff
@@ -0,0 +1,64 @@
...
+                    if maturity not in package['versions']:
+                        continue
+
+                    version = package['versions'][maturity][-1]
+                    pkg.set_version(version.split('-', 1)[0])
+                    pkg.set_rawversion(version)
+                    pkg.set_flags(PackageFlags.devel, maturity == 'test')
+
+                    for subpackage in package['subpackages']:
+                        if subpackage['name'].endswith('-debuginfo'):
```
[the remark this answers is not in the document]
jon-turney: I dropped this, so debuginfo packages are now included.  But I'm not sure how this is supposed to interact with classification (i.e. dragon-debuginfo is not part of dragon-player, but dragon5-part is, somehow, which is confusing to me...)

jon-turney:
Use `rsplit('-', 1)` rather than `split('-', 1)` to split V-R, to handle correctly some antique packages which have a hyphen in the version.

PR #859, opened 2019-06-04
Handle Fedora-style pre-release versions in Cygwin parser

jon-turney:


repology/parsers/parsers/cygwin.py
```diff
@@ -49,9 +50,24 @@ def iter_parse(self, path: str, factory: PackageFactory, transformer: PackageTra
...
-                    verpkg.set_rawversion(version)
+                    raw_version = packagedata['versions'][maturity][-1]
+                    (version, release) = raw_version.rsplit('-', 1)
+
+                    # If release is just '0', that means someone
+                    # forgot it counts from 1, but if it starts with
+                    # '0', the rest indicates the pre-release version
+                    # (as per Fedora/repodata.py)
+                    if release.startswith('0') and len(release) > 1:
+                        match = re.match('0\\.(?:[0-9]+\\.|)((?:alpha|beta|rc)[0-9]+)', release)
```
AMDmi3: Not sure if we need to match the whole structure. Here's all interesting releases from cygwin:

```
0.1
0.1.20151008git
0.1.20171108git
0.1.alpha
0.1.beta1
0.1.rc3
0.1.svn234
0.2
0.2.20150604bzr
0.2.a3
0.2.rc1
0.3
0.3.20180115git
0.3.20180302git
0.4.20180214git
0.6.rc2
0.9.rcgit.29
0.rc7a
```

As far as leading zero is reliable you can just match something like `.*((?:alpha|beta|rc)[0-9]*)`.

Numberless status is usually OK, since it comes from multiple repositories and we handle it anyway. In this case it's `mspack 0.5alpha` and we do handle it for many repos.

More interesting case is `0.2.a3` which belongs to python `3.8.0a3` which is currently also ignored. Since there are no false positives, I'd match this too, taking care to not match it as a part of some unknown suffix by using lookbehind.

It also makes sense to expect `alpha` and the number to be separated, this happens in the wild sometimes.

Finally, I'd match it at the end for safety - all releases follow this rule except for `librsb 1.2.0-0.rc7a`. I don't see such upstream version, so I suppose it should've been something like `1.2.0-0.2.rc7` instead.

Summarizing, I'd do

```python
match = re.fullmatch('.*((?:alpha|beta|rc|(?<=[^a-z])[ab])(\.?[0-9]+)?)', release)
```

jon-turney: Thanks. I've updated the match as you suggested

>  I don't see such upstream version, so I suppose it should've been something like `1.2.0-0.2.rc7` instead.

Yes, that seems to be the case.

jon-turney:
Testing this, there's something not quite right about the capture in some cases.

repology/parsers/parsers/cygwin.py
AMDmi3: > Perhaps the look-behind would be clearer written '?<![a-z]' instead?

Sure.

repology/parsers/parsers/cygwin.py
```diff
@@ -49,9 +50,24 @@ def iter_parse(self, path: str, factory: PackageFactory, transformer: PackageTra
...
+
+                    # If release is just '0', that means someone
+                    # forgot it counts from 1, but if it starts with
+                    # '0', the rest indicates the pre-release version
+                    # (as per Fedora/repodata.py)
+                    if release.startswith('0') and len(release) > 1:
+                        match = re.fullmatch('.*((?:alpha|beta|rc|(?<=[^a-z])[ab])(\.?[0-9]+)?)', release)
+                        if match:
+                            # known pre-release schema
+                            version += '-' + match.group(1)
```
AMDmi3: Just realized, you'd need `verpkg.set_flags(PackageFlags.devel)` here as this is a pre-release version.

jon-turney:
> Testing this, there's something not quite right about the capture in some cases.

Ah, I see what's going on: For '1.2.0-0.rc7a', 'rc7' doesn't match the lookbehind, so 'a' is matched, and we end up extending the version with that as '1.2.0-a'

AMDmi3:
> Ah, I see what's going on: For '1.2.0-0.rc7a', 'rc7' doesn't match the lookbehind, so 'a' is matched, and we end up extending the version with that as '1.2.0-a'

Hmm, this is not right. It would be more correct to require number after `[ab]` then.

Also I've merged conflicts with master, but it did not what I've expected, modifying your branch. Feel free to rebase/squash to avoid confusing history.

AMDmi3:
That is, more like
```regex
.*((?:alpha|beta|rc)(?:\.?[0-9]+)?|(?<![a-z])[ab]\.?[0-9]+)
```

repology/parsers/parsers/cygwin.py
jon-turney: Yeah.  I think maybe that is also missing from repodata.py then, which I copied from...

jon-turney:
Ok, updated match re again, and rebased.

PR #881, opened 2019-07-10
Freshen pkgsrc.

niacat:
- Refer to things as packages instead of ports.
- We're in a stage of migrating away from CVS. Add Mercurial bits.
- Remove old stable branch. I don't think the INDEX file for stable branches is returning.

### DIFF
--- a/repos.d/bsd/pkgsrc.yaml
+++ b/repos.d/bsd/pkgsrc.yaml
@@ -1,40 +1,6 @@
 ###########################################################################
 # pkgsrc
 ###########################################################################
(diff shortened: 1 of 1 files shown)


AMDmi3:
There's no need to have 3 view of the same thing in different repositories, and direct link to Makefile is gone. If mercurial repo is official and working I suggest on only leave it, and a link to Makefile there.

AMDmi3:
Let's just replace CVS links with ones pointing to the new repository when it's officially declared as primary.

PR #897, opened 2019-08-18
Use the `next_page` metadata from crates.io

sgrif:
crates.io will be deprecating its numbered pagination in the near
future. In order to help clients migrate to the new pagination API, a
`next_page` and `prev_page` field is included in the response metadata,
unless you're on the first/last page.

Currently this should not result in any change in functionality (the
values in `next_page` will equal what was being hit before), but once
the new pagination system is in place, this field will switch to using
it, meaning Repology will be unaffected when the old pagination is
eventually removed.

I was not able to test this change locally, as there doesn't seem to be
a way to just run a single fetcher without also having a database set
up, installing C libraries that aren't published to any package manager
that I could find, etc.

Thanks for setting a useful user-agent header so I can reach out with
this PR instead of just breaking your app

### DIFF
--- a/repology/fetchers/fetchers/cratesio.py
+++ b/repology/fetchers/fetchers/cratesio.py
@@ -33,7 +33,7 @@ class CratesIOFetcher(ScratchDirFetcher):
 
     def _do_fetch(self, statedir: AtomicDir, persdata: PersistentData, logger: Logger) -> bool:
         numpage = 1
-        query ='?per_page={}&sort=alpha'.format(self.per_page)
+        query = '?per_page={}&sort=alpha'.format(self.per_page)
         while query:
             url = self.url + query
             logger.log('getting ' + url)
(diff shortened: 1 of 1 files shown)


AMDmi3:
Thanks a lot for this submission! This is broken though as `next_page` field only contains URL parameters starting with `?`, not the full URL, so it should still be concatenated with `self.url`. Can be fixed with e.g.

```diff
-            # parse next page
-            url = json.loads(text)['meta']['next_page']
+            next_page = json.loads(text)['meta'].get('next_page')
+            if not next_page:
+                logger.log('last page detected')
+                break
+
+            url = self.url + next_page
```

Care to fix it? You could also add yourself to the copyright header at the top of the file.

sgrif:
Ah thanks, I forgot that the URL is relative to the current page. I've fixed the code. I did not add myself to the copyright header, I am fine with transferring the copyright of this code to the project owners.

AMDmi3:
Merged, thank you! Taking the opportunity, I'd mention that a JSON dump of all crates.io package metadata would be very welcome.

sgrif:
It's in the works. https://github.com/rust-lang/crates.io/issues/630

PR #918, opened 2019-09-19
Updated the pureos.yaml file

jeremiah:
Hi - thanks for repology! I'm updating our PureOS release name so you have an accurate package count.

### DIFF
--- a/repos.d/deb/pureos.yaml
+++ b/repos.d/deb/pureos.yaml
@@ -1,9 +1,9 @@
 ###########################################################################
 # PureOS
 ###########################################################################
-- name: pureos_green
+- name: pureos_amber
   type: repository
-  desc: PureOS green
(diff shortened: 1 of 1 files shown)


AMDmi3:
Thanks!

However, what's the difference between green and amber? Shouldn't it be added instead of replacing? Is landing still actual?

jeremiah:
Thank you!

The difference is that 'green' was following Debian testing. We're going to turn 'green' into two distros, one following Debian stable and one following Debian testing. We put our packages on top but we try and keep a very small delta between us and Debian. We've created the code name amber for the stable and byzantium for the testing release. Amber is the one that's ready now.

jeremiah:
Oh, yes, landing is still current (actual.) :+1:

AMDmi3:
Thanks for clarification, looks good.

Issue #925, opened 2019-09-28
Dropping maintainer info from Funtoo


juippis:
Hey, 

I'm kindly approaching due to Funtoo "wrongly" reporting the package status for maintainers. I saw a recent commit about disabling maintainers from certain repos, 
5ef34e63ab04777b7503447335074f6a5f6a58ee
and believe Funtoo should be added to that list. Gentoo developers have no control over what gets pulled to Funtoo and when, so ever since Funtoo became a release-based distribution, it falsifys up-to-date status for Gentoo developers in Repology. Also as @palica said, https://github.com/repology/repology/issues/824#issuecomment-481530759 Funtoo doesn't have a maintainer concept - they just copy metadata from Gentoo. 

So I'm kindly asking to drop maintainer info from Funtoo in Repology as I see it's possible to do (I can make a PR about it), 
or alternatively @danielrobbins to update all package maintainer info into "@funtoo.org" (might require a bit more work).

I invite discussion about this subject, I know many Gentoo developers have shown frustration about this.

danielrobbins:
I do not use repology so I don't know what the "fuss" is about, but it seems like having Funtoo make changes to metadata for the sake of repology is a bit like the tail wagging the dog.

As far as this issue -- all I can say is that I am totally unware of this until it was brought to my attention by @juippis in #funtoo on freenode just now. A comment from @palica, a Funtoo user, doesn't indicate that this issue has ever been explained to the Funtoo project. So it is completely off our radar. Therefore, any mitigation on repology's end seems premature.

I would appreciate anyone concerned about this in regards to funtoo to OPEN A BUG on the funtoo bug tracker so the exact problem can be described in detail and a solution proposed.

danielrobbins:
Regarding the yet-to-be discussed technical issue, Funtoo takes a snapshot of Gentoo at a particular point in time. Most of the ebuilds in Funtoo come from that snapshot, and are maintained by those maintainers that metadata.xml says they are. So I don't see what the issue is here. Please describe the actual problem in detail so I can understand it.

[...]

Issue #655, opened 2018-07-10
Returns packages that don't satisfy search criteria


yurivict:
This query
https://repology.org/metapackages/?search=&maintainer=yuri%40freebsd.org&category=&inrepo=freebsd&notinrepo=&repos=&families=&repos_newest=&families_newest=&outdated=on
should return all outdated FreeBSD packages maintained by me.

It returns python:pycryptodome which is outdated on FreeBSD but isn't maintained by me.

AMDmi3:
True, because repository and maintainer conditions are checked independently, so it's shown because you are still a maintainer in dports. This specific problem will go away as soon as dports are updated, to properly fix it some database changes would be required.

mgorny:
Bump. I'm also having major problems due to this bug. It seems that Funtoo (a Gentoo fork) includes a lot of Gentoo packages with outdated maintainer information. Then, e.g. when I'm searching for Gentoo packages maintained by `proxy-maint@gentoo.org` I'm getting a lot of results for packages that were maintained by us in the past.

AMDmi3:
I guess I have to force this, as there are repeated reports from at least Gentoo maintainers affected by this issue.

A short explanation of the problem and why it's not fixed yet. The problem is that *In repo* and *Maintainer* conditions in the project search are checked independently, so if you try to list packages maintainer by _foo@gentoo.org_ in _Gentoo_ it lists projects with have packages maintained by _foo@gentoo.org_ AND packages in _Gentoo_, while it should list projects with packages in _Gentoo_ AND maintained by _foo@gentoo.org_ at the same time.

Fixing this is not quite straightforward due to how Repology database is designed and may make the query too slow and/or require additional tables with aggregate information which would slow down the update and eay the diskspace. So I've postponed this until long planned update process refactoring.

Since latter is unlikely to be ready soon, I'm starting to work on fixing this with the current codebase, even if there are (tolerable) drawbacks.

AMDmi3:
Btw, examples of this behaviour would be helpful. I have one affecting myself with FreeBSD/DPorts, but it may go away as soon as DPorts are updated.

mgorny:
Example: https://repology.org/projects/?inrepo=gentoo&maintainer=proxy-maint%40gentoo.org

The first package on the list, 389-ds-base is reported wrongly.

AMDmi3:
[...]

PR #926, opened 2019-09-29
Patch 2

palica:
please update Funtoo to 1.4 release, this should also fix the problem with metadata.xml showing incorrect information about funtoo vs gentoo.

https://github.com/repology/repology/issues/925

### DIFF
--- a/repos.d/gentoo/funtoo.yaml
+++ b/repos.d/gentoo/funtoo.yaml
@@ -80,7 +80,108 @@
       subrepo: '{source}'
   repolinks:
     - desc: Funtoo Linux home
-      url: http://www.funtoo.org/Welcome
+      url: https://www.funtoo.org
+    - desc: GitHub organization
+      url: https://github.com/funtoo/
+  packagelinks:
+    - desc: Git directory
+      url: 'https://github.com/funtoo/{subrepo}/tree/master/{category}/{name}'
+    - desc: View ebuild
+      url: 'https://github.com/funtoo/{subrepo}/blob/master/{category}/{name}/{name}-{rawversion}.ebuild'
+  tags: [ all, production, funtoo ]
+
+# Funtoo - 1.4 Release (Sept 11 2019)
+- name: funtoo_1.4
+  type: repository
+  desc: Funtoo 1.4
+  family: gentoo
+  ruleset: [gentoo, funtoo]
+  color: '4e4e71'
+  minpackages: 35000
+  default_maintainer: fallback-mnt-funtoo@repology
+  sources:
+    - name:
+        - core-gl-kit
+        - core-hw-kit
+        - core-kit
+        - core-server-kit
+        - core-ui-kit
+        - desktop-kit
+        - dev-kit
+        - editors-kit
+        - games-kit
+        - haskell-kit
+        - java-kit
+        - kde-kit
+        - lang-kit
+        - lisp-scheme-kit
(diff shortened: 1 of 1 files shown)


AMDmi3:
I won't accept dropping metadata. Homepage information is crucial for spitting conflicting projects.

danielrobbins:
@AMDmi3 homepage information should be coming from the md5-cache or the ebuild, it is not stored in the metadata.It is fine to drop the metadata, I checked your parser source code.

AMDmi3:
Homepage also includes information from `<upstream>`.

danielrobbins:
@AMDmi3 OK, well I can do a patch so that there is a way through the YAML to disable parsing of just the maintainer info from the metadata.xml. Would that be accepted by you? If so, can you provide info on how you would like it to work and I will submit the patch.

AMDmi3:
The point is that you don't need do hide Funtoo maintainers at all for Repology, see #925 amd #655.

danielrobbins:
OK, gotcha. We can re-send the patch with the metadata.xml still in it as it contains updates for Funtoo 1.4.

I suspect this issue is what is described in #655.

repos.d/gentoo/funtoo.yaml
```diff
@@ -2,45 +2,51 @@
...
-        - ml-lang-kit
-        - net-kit
-        - nokit
+        - dev-kit
         - python-modules-kit
-        - ruby-kit
-        - science-kit
-        - security-kit
-        - text-kit
+        - nokit
```
AMDmi3: Please keep this sorted to reduce diff

repos.d/gentoo/funtoo.yaml
```diff
@@ -2,45 +2,51 @@
...
+        - net-kit
+        - text-kit
+        - science-kit
         - games-kit
-        - haskell-kit
+        - perl-kit
         - java-kit
-        - kde-kit
-        - lang-kit
+        - ruby-kit
```
AMDmi3: You must not duplicate repositories which we handle separately (due to different branch names) below here.

AMDmi3:
Also, does it fully obsolete 1.3? If there's a change that some people still use 1.3, we may leave it there for informational purpose.

palica:
according to this you should be able to squash when merging. https://github.blog/2016-04-01-squash-your-commits/

palica:
this is the documentation for it https://help.github.com/en/articles/about-pull-request-merges

AMDmi3:
> you should be able to squash when merging

I'd prefer to merge properly prepared commit as is. Squashing via GH requires me to edit the commit message and it may end up not what a contributor expects. Also I suspect it would overwrite committer field in the commit.

repos.d/gentoo/funtoo.yaml
palica: should be fixed now

repos.d/gentoo/funtoo.yaml
palica: should be fixed now

palica:
rebase done

AMDmi3:
Ok, looks good. So what about leaving 1.3 be for now? @danielrobbins?

danielrobbins:
@AMDmi3 if you have to choose one, choose 1.4, since that is the current release, but if you can include both then it is fine to include 1.3 since it is still supported. Thanks.

PR #934, opened 2019-10-16
Remove getdeb/playdeb

patrick96:
I removed the repos because they don't exist anymore and clutter up the package list. For example we have five getdeb packages in the list for our project [polybar](https://repology.org/project/polybar/versions), which will never get updated and can't even be downloaded.

If I understood your comment in the YAML file correctly, you only kept the repo as a reference point for projects worth packaging.

However the full list of getdeb and playdeb packages (buid files and everything) are available on the original getdeb maintainer's github ckorn:

https://github.com/ckorn/GetDeb

https://github.com/ckorn/PlayDeb

If I misunderstood the reason why you kept the repos and you still want to keep them, feel free to close this.


AMDmi3:
You are right, it's time to remove this.

PR #945, opened 2019-11-08
Start tracking KISS testing repository

dylanaraps:
It no longer makes any sense to ignore this repository. Please let me know if any further changes need to be done. :)

**Note**: Two parsers have the execute permission (**kiss** and **fink**) while all others don't. Shall I update the permissions?


AMDmi3:
> Two parsers have the execute permission (kiss and fink) while all others don't. Shall I update the permissions?

No, it's a bug.

PR #961, opened 2019-12-19
add repodata db support

jmahler:
The repodata in an RPM repo can contain data in both xml and sqlite
formats.  Currently Repodata{Parser,Fetcher} only supports
the xml format.  However, in some cases xml may be missing or it
sqlite might be faster.

Add support for sqlite db formats using new RepodataDb{Parser,Fetcher} classes.

### Testing

Switching to RepodataDb{Fetcher,Parser} works with an existing repo.

```
osdev:repology-updater$ git diff
diff --git a/repos.d/rpm/centos.yaml b/repos.d/rpm/centos.yaml
index f29f2427..0b72f99b 100644
--- a/repos.d/rpm/centos.yaml
+++ b/repos.d/rpm/centos.yaml
@@ -33,8 +33,8 @@
   minpackages: 3000
   sources:
     - name: [ os, updates, extras ]
-      fetcher: RepodataFetcher
-      parser: RepodataParser
+      fetcher: RepodataDbFetcher
+      parser: RepodataDbParser
       url: 'http://vault.centos.org/centos/7/{source}/Source/'
       subrepo: '{source}'
   repolinks:
osdev:repology-updater$ sudo -u repology ./repology-update.py --fetch --fetch --parse --repositories centos_7
Dec 19 16:26:10 loading rules
Dec 19 16:26:13 updating repositories metadata
Dec 19 16:26:13   committing changes
Dec 19 16:26:13 fetching centos_7
Dec 19 16:26:16   done
Dec 19 16:26:27 parsing centos_7
Dec 19 16:26:34   done
Dec 19 16:26:34 total time taken: 23.97 seconds
```

Old Repodata{Fetcher,Parser} still work.

```
osdev:repology-updater$ sudo -u repology ./repology-update.py --fetch --fetch --parse --repositories centos_6
Dec 19 16:28:19 loading rules
Dec 19 16:28:22 updating repositories metadata
Dec 19 16:28:22   committing changes
Dec 19 16:28:22 fetching centos_6
Dec 19 16:28:24   done
Dec 19 16:28:35 parsing centos_6
Dec 19 16:28:42   done
Dec 19 16:28:42 total time taken: 23.45 seconds
```

### DIFF
--- a/repology/fetchers/fetchers/repodata.py
+++ b/repology/fetchers/fetchers/repodata.py
@@ -25,6 +25,8 @@ from repology.logger import Logger
 
 
 class RepodataFetcher(ScratchFileFetcher):
+    primary_key = 'primary'
+
     def __init__(self, url: str, fetch_timeout: int = 60):
         super(RepodataFetcher, self).__init__(binary=True)
 
@@ -32,14 +34,15 @@ class RepodataFetcher(ScratchFileFetcher):
         self.fetch_timeout = fetch_timeout
 
     def _do_fetch(self, statefile: AtomicFile, persdata: PersistentData, logger: Logger) -> bool:
+
         # fetch and parse repomd.xml
         repomd_url = self.url + 'repodata/repomd.xml'
         logger.log('fetching metadata from ' + repomd_url)
         repomd_content = do_http(repomd_url, check_status=True, timeout=self.fetch_timeout).text
         repomd = xml.etree.ElementTree.fromstring(repomd_content)
-        repomd_elt_primary = repomd.find('{http://linux.duke.edu/metadata/repo}data[@type="primary"]')
+        repomd_elt_primary = repomd.find('{{http://linux.duke.edu/metadata/repo}}data[@type="{}"]'.format(self.primary_key))
         if repomd_elt_primary is None:
-            raise RuntimeError('Cannot find <primary> element in repomd.xml')
+            raise RuntimeError('Cannot find <{}> element in repomd.xml'.format(self.primary_key))
 
         repomd_elt_primary_location = repomd_elt_primary.find('./{http://linux.duke.edu/metadata/repo}location')
         repomd_elt_primary_checksum = repomd_elt_primary.find('./{http://linux.duke.edu/metadata/repo}open-checksum[@type="sha256"]')
@@ -61,6 +64,8 @@ class RepodataFetcher(ScratchFileFetcher):
             compression = 'gz'
         elif repodata_url.endswith('xz'):
             compression = 'xz'
+        elif repodata_url.endswith('bz2'):
+            compression = 'bz2'
 
         logger.log('fetching {}'.format(repodata_url))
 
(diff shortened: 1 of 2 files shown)


repology/fetchers/fetchers/repodata.py
```diff
@@ -25,21 +25,28 @@
 
 
 class RepodataFetcher(ScratchFileFetcher):
-    def __init__(self, url: str, fetch_timeout: int = 60):
+    def __init__(self, url: str, fetch_timeout: int = 60, use_db: bool = False):
         super(RepodataFetcher, self).__init__(binary=True)
 
         self.url = url
         self.fetch_timeout = fetch_timeout
+        self.use_db = use_db
```
AMDmi3: I'd instead set `primary_key = 'primary'` as the class member. Derived class would then override it to `primary_db`. Constructor argument goes away (which is good as it would allow confusing uses), and you don't even need to override constructor in the derived class.

here and remove the constructor argument (which may lead to confusing uses). `RepodataDbFetcher` constructor would set it to `primary_db`

repology/fetchers/fetchers/repodata.py
```diff
@@ -25,21 +25,28 @@
...
 
     def _do_fetch(self, statefile: AtomicFile, persdata: PersistentData, logger: Logger) -> bool:
+
```
AMDmi3: this goes away and later you just use `self.primary_key`

repology/fetchers/fetchers/repodatadb.py
```diff
@@ -0,0 +1,24 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with repology.  If not, see <http://www.gnu.org/licenses/>.
+
+from repology.fetchers.fetchers.repodata import RepodataFetcher
+
+
+class RepodataDbFetcher(RepodataFetcher):
```
AMDmi3: Let's declare this after the `RepodataFetcher` in the same file. They are logically grouped.

Also I'd name it `RepodataSqliteFetcher` as more meaningful.

repology/parsers/parsers/repodatadb.py
```diff
@@ -0,0 +1,47 @@
...
+
+from repology.packagemaker import NameType, PackageFactory, PackageMaker
+from repology.parsers import Parser
+from repology.parsers.nevra import nevra_construct
+from repology.parsers.sqlite import iter_sqlite
+from repology.parsers.versions import VersionStripper
+from repology.transformer import PackageTransformer
+
+
+class RepodataDbParser(Parser):
```
AMDmi3: Likewise, let's move this into the same file with `RepodataParser` and name it `RepodataSqliteParser`.

repology/fetchers/fetchers/repodata.py
jmahler: Great suggestions, this definitely cleans it up.  Fixing in v2 ...

jmahler:
I believe I addressed all the comments.  And code coverage should improve once a source that uses RepodataSqlite{Parser,Fetcher} is added.

Let me know if it needs anything else.

AMDmi3:
I've fixed maintainer extraction (extract_maintainer is needed to get just email) and version handling (same way as it's done in the old parser; interestingly this helped to uncover a bug there). I haven't looked deep into sqlite format, could you give some examples of which information can be available there but not in XML format?

Issue #960, opened 2019-12-19
Package in Exherbo app-misc not showing


jarun:
A user brought to my notice that `nnn` is available in Exherbo under the app-misc category:

https://git.exherbo.org/summer/packages/app-misc/nnn/index.html

However, it doesn't show in the repology badge: https://repology.org/badge/vertical-allrepos/nnn.svg

Can you please fix this?

AMDmi3:
You may want to refresh the repository list at https://github.com/repology/repology-updater/blob/master/repos.d/exherbo.yaml

jarun:
> You may want to refresh

How do I do that?

jarun:
Do you mean some sections like `app-misc` are missing

jarun:
OK, got the package and the dev. Thanks for the pointer.

AMDmi3:
As I understand exherbo is a combination of a bunch of repositories. Currently, repos listed on https://git.exherbo.org/ are used. nnn comes from https://gitlab.exherbo.org/DanySpin97/danyspin97-exheres not listed anywhere. So we need an actual listing of repositories Exherbo consists of, and Repology config needs to be updated with it.

jarun:
Raised PR #962.

DanySpin97:
Hi there. I am the mantainer of nnn package in Exherbo.

> As I understand exherbo is a combination of a bunch of repositories.

Yup, there are 2 types of repositories: official ones (both Exherbo repositories and developer's personal ones) and unofficial ones. The official ones are already included by repology.

About the unofficial ones; they are (strictly) reviewed by Exherbo devs and then added to [this folder](https://gitlab.exherbo.org/exherbo-misc/infra-scripts/tree/master/playboy/exherbo-unofficial/config/repositories). They do not contain duplicate packages.

The format is the same, but the repositories can be anywhere; almost all of them are on gitlab.exheres.org but some are on github.com too. In the next days I will have a look at the way repology parse [exherbo.yaml](https://github.com/repology/repology-updater/blob/master/repos.d/exherbo.yaml) and add the unofficial ones too.

AMDmi3:
Isn't there a complete merged information on all supported packages? Maybe somewhere around https://git.exherbo.org/summer/packages/app-misc/nnn/index.html?

There are two problems with current approach:
- We have to maintain a list of Exherbo repositories in Repology, which is duplication of something already done in exherbo itself, and will always introduce lag and lead to issues like this one.
[...]

PR #962, opened 2019-12-19
Add source dev/danyspin97

jarun:
Ref: issue #960.

### DIFF
--- a/repos.d/exherbo.yaml
+++ b/repos.d/exherbo.yaml
@@ -50,6 +50,7 @@
         - dev/ciaranm
         - dev/cogitri
         - dev/compnerd
+        - dev/danyspin97
         - dev/dleverton
         - dev/ferdy
         - dev/heirecka


AMDmi3:
There's no such repository. Since it comes from outside of git.exherbo.org it needs a special handling. But as suggested in #960 let's start from getting complete list of exherbo repositories, as others may be missing. And/or there may by a way to get all the repositories in one go I don't know about.

jarun:
I see one difference.

For this unlisted repo the status is marked as third party: https://gitlab.exherbo.org/DanySpin97/danyspin97-exheres/blob/master/metadata/about.conf

For a listed repo, the status is dev. For example:

https://gitlab.exherbo.org/exherbo-devs/zlin/blob/master/metadata/about.conf

jarun:
And as you mentioned, the repository is not linked as well.

AMDmi3:
I have no idea of how the exherbo repositories are classified. If the package shows up on https://git.exherbo.org/summer/packages/ and/or in the system, there should be a list of all included repositories somewhere.

PR #965, opened 2019-12-27
Add Amazon Linux distribution

jmahler:
Adds the Amazon Linux distribition (both AL1 and AL2).

### Testing

Ran repology-update.py and confirmed results in repology-webapp looks reasonable.

```
alosdev:repology-updater$ sudo -u repology ./repology-update.py --fetch --fetch --parse --database --postupdate --repositories amazonlinux
```

### DIFF
--- a/repos.d/rpm/amazonlinux.yaml
+++ b/repos.d/rpm/amazonlinux.yaml
@@ -1,0 +1,30 @@
+- name: amazon_2
+  type: repository
+  desc: Amazon Linux 2
+  family: fedora
+  minpackages: 2000
+  sources:
+    - name: core
+      fetcher: RepodataSqliteFetcher
+      parser: RepodataSqliteParser
+      url: http://amazonlinux.us-west-2.amazonaws.com/2/core/2.0/SRPMS/mirror.list
+  repolinks:
+    - desc: Amazon Linux 2
+      url: https://aws.amazon.com/amazon-linux-2/
+  tags: [ all, production, rpm, amazon ]
+
+- name: amazon_1
+  type: repository
+  desc: Amazon Linux 1
+  family: fedora
+  minpackages: 2000
+  sources:
+    - name: [ main, updates ]
+      fetcher: RepodataSqliteFetcher
+      parser: RepodataSqliteParser
+      url: 'http://repo.us-west-2.amazonaws.com/latest/{source}/SRPMS/mirror.list'
+      subrepo: '{source}'
+  repolinks:
+    - desc: Amazon Linux AMI
+      url: https://aws.amazon.com/amazon-linux-ami/
+  tags: [ all, production, rpm, amazon ]


jmahler:
Was failing because minpackages wasn't set.  Set to 2000.  Now it works.

```
alosdev:repology-updater$ ./repology-schemacheck.py -s repos repos.d/rpm/amazonlinux.yaml 
```
OK

repos.d/rpm/amazonlinux.yaml
```diff
@@ -0,0 +1,33 @@
+- name: amzn2
```
AMDmi3: I'd avoid unnecessary abbreviation and stick to how other repos are named, so probably `amazon_1` and `amazon_2`

repos.d/rpm/amazonlinux.yaml
```diff
@@ -0,0 +1,33 @@
...
+      parser: RepodataSqliteParser
+      url: http://amazonlinux.us-west-2.amazonaws.com/2/core/2.0/SRPMS/mirror.list
+  repolinks:
+    - desc: Amazon Linux 2
+      url: https://aws.amazon.com/amazon-linux-2/
+  tags: [ all, production, rpm, amazonlinux ]
+
+- name: amzn1
+  type: repository
+  desc: Amazon Linux 1
```
AMDmi3: Is it "1" or "AMI"? It feels like one of them should be used in both desc and repolinks.

repos.d/rpm/amazonlinux.yaml
```diff
@@ -0,0 +1,33 @@
...
+  minpackages: 2000
+  sources:
+    - name: core
+      fetcher: RepodataSqliteFetcher
+      parser: RepodataSqliteParser
+      url: http://amazonlinux.us-west-2.amazonaws.com/2/core/2.0/SRPMS/mirror.list
+  repolinks:
+    - desc: Amazon Linux 2
+      url: https://aws.amazon.com/amazon-linux-2/
+  tags: [ all, production, rpm, amazonlinux ]
```
AMDmi3: Just `amazon` here, as it's mostly a shortcut for command line and should not be much different from repository names.

repos.d/rpm/amazonlinux.yaml
```diff
@@ -0,0 +1,33 @@
+- name: amzn2
+  type: repository
+  desc: Amazon Linux 2
+  family: fedora
```
AMDmi3: I'd like a comment on that: is it known to follow fedora package naming style (e.g. not centos/mandriva/mageia/suse/...)? Maybe it's known to be based on specific distro?

repos.d/rpm/amazonlinux.yaml
```diff
@@ -0,0 +1,33 @@
...
+  type: repository
+  desc: Amazon Linux 1
+  family: fedora
+  minpackages: 2000
+  sources:
+    - name: updates
+      fetcher: RepodataSqliteFetcher
+      parser: RepodataSqliteParser
+      url: http://repo.us-west-2.amazonaws.com/latest/updates/SRPMS/mirror.list
+    - name: main
```
AMDmi3: Both sources can be described by one entry, with `- name: [main, updates]`, `url: 'http://repo.us-west-2.amazonaws.com/latest/{source}/SRPMS/mirror.list'` and added `subrepo: '{source}'` see rpm/centos.yaml for example.

jmahler:
> Isn't there a place where .spec's can be viewed, e.g. to produce packagelinks?

You mean like this link for Fedora?

https://src.fedoraproject.org/rpms/vim

Unfortunately there isn't currently an equivalent for Amazon Linux.
Only the binary and source packages are provided via the repos.

repos.d/rpm/amazonlinux.yaml
jmahler: AMI is more consistent with the links, will fix in v2

repos.d/rpm/amazonlinux.yaml
jmahler: it is closest to centos though some packages are derived from fedora.  Will switch to centos.

jmahler:
### Changes in v2
- Addressed all comments except for the one about package links which are not currently available.

repos.d/rpm/amazonlinux.yaml
jmahler: On second thought ... "Amazon Linux 1" and "Amazon Linux AMI" can be used interchangeably.  But "Amazon Linux 1" is more clear when listed by "Amazon Linux 2".  Will switch back main desc to "Amazon Linux 1" but will leave repo link desc to "Amazon Linux AMI".

AMDmi3:
Let's switch back to `family: fedora`, the test I've run overnight has shown that using centos introduces some pessimizations. With that it would be good to go.

PR #997, opened 2020-01-28
Add native DPorts git, point to it by default

mikhailnov:


AMDmi3:
Let's order all DPorts specific urls first then - `DPorts git`, `DPorts Makefile`, `Synth build log`, then all FreeBSD links.

mikhailnov:
done

PR #1021, opened 2020-03-28
Add homepage parsing support for Cygwin

jon-turney:
This information is now propagated for new packages.  I'll also be working on backfilling it for existing packages.

### DIFF
--- a/repology/parsers/parsers/cygwin.py
+++ b/repology/parsers/parsers/cygwin.py
@@ -41,6 +41,9 @@ class CygwinParser(Parser):
                 if 'maintainers' in packagedata:
                     pkg.add_maintainers([m.replace('.', '').replace(' ', '.') + '@cygwin' for m in packagedata['maintainers']])
 
+                if 'homepage' in packagedata:
+                    pkg.add_homepages(packagedata['homepage'])
+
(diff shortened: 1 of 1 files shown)


AMDmi3:
Cool, thanks! It would be nice to have homepages for
- alternatives
- gdl
- libusb
- nc
- pal
- rebase
- screenruler
- signify
- words
in cygwin for it would allow us to get rid of the bunch of rules.

jon-turney:
I've backfilled homepage data so it should now be available for all our packages which record it.  I'll look at adding it to those packages on your list where it's still missing.

 - [x] alternatives (http://www.fastcoder.net/software/chkconfig/)
 - [ ] gdl
 - [x] libusb (libusb-win32: http://libusb-win32.sourceforge.net/, libusb1.0: https://libusb.info/)
 - [x] nc (http://www.openbsd.org/cgi-bin/cvsweb/src/usr.bin/nc)
 - [ ] pal
 - [ ] rebase
 - [x] screenruler (https://gnomecoder.wordpress.com/screenruler/)
 - [ ] signify
 - [x] words (https://en.wikipedia.org/wiki/Moby_Project)

PR #1029, opened 2020-04-17
Fix Fedora 31 release name

dreua:
I think that this only changes the displayed name from development to release, if it has any unwanted side effects this should probably not be merged. 

With the change the naming matches the naming scheme of the other Fedora repositories in repology and of course the actual purpose of this repository (it contains the packages of the release and has not been updated since then).

### DIFF
--- a/repos.d/rpm/fedora.yaml
+++ b/repos.d/rpm/fedora.yaml
@@ -185,7 +185,7 @@
   color: '294172'
   minpackages: 20000
   sources:
-    - name: development
+    - name: release
       fetcher: RepodataFetcher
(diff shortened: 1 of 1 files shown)


AMDmi3:
It doesn't change anything user visible - it's internally used name which affects the name of directory for downloaded state. Still, it's correct change.

dreua:
Actually I hoped this would change this text:
repology_packages_draw


But I'm happy to improve your internal download names as well :smiley:

AMDmi3:
Oh, I forgot about that. Yes, that'll change too.

PR #1038, opened 2020-05-05
Allow submodules

jedahan:
Would like to start the conversation so we can see the entire kiss universe of packages

### DIFF
--- a/repology/fetchers/fetchers/git.py
+++ b/repology/fetchers/fetchers/git.py
@@ -55,13 +55,14 @@ class GitFetcher(PersistentDirFetcher):
             os.fsync(sparse_checkout_file.fileno())
 
     def _do_fetch(self, statepath: str, logger: Logger) -> bool:
-        Runner(logger=logger).run('timeout', self._timeout_arg, 'git', 'clone', '--progress', '--no-checkout', self._depth_arg, '--branch', self._branch, self._url, statepath)
+        Runner(logger=logger).run('timeout', self._timeout_arg, 'git', 'clone', '--recursive', '--progress', '--no-checkout', self._depth_arg, '--branch', self._branch, self._url, statepath)
 
         r = Runner(logger=logger, cwd=statepath)
 
         r.run('git', 'config', 'core.sparsecheckout', 'true')
         self._setup_sparse_checkout(statepath)
         r.run('git', 'checkout')
+        r.run('timeout', self._timeout_arg, 'git', 'submodule', 'update', '--init', '--recursive')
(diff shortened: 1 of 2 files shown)


AMDmi3:
- Submodules do not properly support sparse checkout here, and it looks like some _do_update code is also lacking. Not sure if the complication needed to properly support it would be acceptable.
- Repology policy is against supporting individual repositories or collections thereof. Also, I'll never allow something like kiss-dbus which includes dbus sources.

jedahan:
So from what I understand

- [ ] figure out what it means to properly support sparse checkout
  - does this mean only checking out submodules if they exist? or is there some other specific issue you mean?
- [ ] add support for `_do_update`
- [ ] remove kiss-dbus from kiss-universe until they can separate the packages from the source

What I'm confused by is "Repology policy is against supporting individual repositories or collections thereof". Aren't all collections of packages some sort of repository or collection thereof? Or do you mean having both commits (one for submodules, one for updating kiss) should be split into two PRs? (I just added the second commit to illustrate a test case)

jedahan:
Ohh, or do you mean it being under my personal repo, `jedahan` - if so I am talking with the kiss maintainers to see if there's a more appropriate place.

AMDmi3:
> does this mean only checking out submodules if they exist? or is there some other specific issue you mean?

I suspect that you'd need to manually iterate through submodules recursively and setup sparse checkouts for them, and repeat that on each update to handle addition and removal of submodules. This already looks too complex and likely slow for the purpose.

> remove kiss-dbus from kiss-universe until they can separate the packages from the source

I do not want Repology to affect your repository contents, just saying that having to move whole sourcetrees around just for the purpose of ignoring them is not acceptable for Repology.

You may consider parsing repositories on your side to produce ready to use machine readable metadata dump - it would simplify things on both sides IMO.  

> What I'm confused by is "Repology policy is against supporting individual repositories or collections thereof".

I've meant to say _personal_, not _individual_. Personal as in maintained by a single person as opposed to community.

dylanaraps:
It's worth noting that the only official repositories (managed by me) are kisslinux/repo (Official repositories) and kisslinux/community (Community repository). Repology is tracking KISS Games (and maybe others, haven't looked). Anything other than the repo and community are unofficial.

Not sure how you want to handle this. I see no point in including anything outside of KISS' official repositories as there's nowhere you can draw the line (there are 50~ user/unofficial repositories and this number is growing by the day).

AMDmi3:
We don't want to support submodules.

PR #1044, opened 2020-05-14
Add repository KISS Games

IslandAlpha:
As this is a community repository, it is best left in it's own file rather than in kiss.yaml.

It may be better to put kiss.yaml and kiss-games.yaml in their own folder, if this works let me know.


AMDmi3:
Merged

Issue #1054, opened 2020-05-28
extract_maintainers doesn't work for CRUX


TimB87:
Hi!
I wish my python skills were better so I could fix that myself, but time currently permits me to dig deeper into it.

From my understanding, extract_maintainers function doesn't work with the crux way of declaring Maintainers in Pkgfiles (example: [dash Pkgfile](https://crux.nu/gitweb/?p=ports/core.git;a=blob;f=dash/Pkgfile;h=c2e616b99900b59cc5970cb9d4e431cba8c8f04d;hb=HEAD)):
`# Maintainer: (first name) name, email at host dot tld`

Is it easily possibly to get this going?

AMDmi3:
By using non-obfuscated emails in CRUX, yes.
In Repology - no, we purposely do not and will not deobfuscate emails.

TimB87:
@AMDmi3 I'll close the issue as is as there is nothing to be done then (not my decision, not sure how the CRUX team feels about that either).

Anyway, thanks for the quick reply and repology as a whole! 👍

AMDmi3:
Couple of additional thoughts:
- I don't think CRUX team would force you to obfuscate _your_ email if you don't want to
- At some point, CRUX parsing in Repology would anyway have to either be switched from parsing Pkgfiles as plaintext (incomplete and unreliable) to parsing some kind of machine readable dump of package metadata produced on CRUX side, or be removed. When/if such a dump is implemented, emails can be converted to readable form in the process of generating it (probably on opt-in basis per maintainer). Ataraxia linux uses package format similar to CRUX and have developed a way to generate such a dump (#913), which probably can be reused for CRUX. Just telling if you or someone else from CRUX community would be interesting in keeping/improving Repology support in future.

TimB87:
Thanks for the valuable input!
In the long term I would prefer to keep repology going for us as well, so just unobfuscating my mails would solve anything. I'll try and talk to people and see what we can get going here!

AMDmi3:
Still closing as there's nothing to do here on Repology side.

PR #1058, opened 2020-06-11
Updated CRUX parser class and repository config to use a JSON dump ra…

mhoush:
…ther than the deprecated plaintext parser

This hopefully fixes issue #1054 and makes for a more future-proof setup. I'm new to repology so it's certainly possible this can be improved. Comments/questions welcome, of course.

With that said, I removed the 'statsgroup' item from the repo config as it appeared to only make sense in the context of multiple repositories in the same config (such as the old separation between CRUX versions). Only the current version is supported for the most part. This is also the reason for the change of name and description.

Tagging @TimB87 as well.

### DIFF
--- a/repology/parsers/parsers/crux.py
+++ b/repology/parsers/parsers/crux.py
@@ -15,68 +15,27 @@
 # You should have received a copy of the GNU General Public License
 # along with repology.  If not, see <http://www.gnu.org/licenses/>.
 
-import os
 from typing import Iterable
 
 from repology.logger import Logger
 from repology.packagemaker import NameType, PackageFactory, PackageMaker
 from repology.parsers import Parser
 from repology.parsers.maintainers import extract_maintainers
+from repology.parsers.json import iter_json_list
 from repology.transformer import PackageTransformer
 
-
-class CRUXParser(Parser):
+class CRUXPortsJsonParser(Parser):
     def iter_parse(self, path: str, factory: PackageFactory, transformer: PackageTransformer) -> Iterable[PackageMaker]:
-        for pkgdir in os.listdir(path):
-            pkgpath = os.path.join(path, pkgdir, 'Pkgfile')
-            if not os.path.exists(pkgpath):
-                continue
-
-            with open(pkgpath, 'r', encoding='utf-8', errors='ignore') as pkgfile:
-                pkg = factory.begin()
-
-                name = None
-                version = None
-
-                for line in pkgfile:
-                    line = line.strip()
-                    if line.startswith('# Description:'):
-                        pkg.set_summary(line.split(':', 1)[1])
-
-                    if line.startswith('# URL:'):
-                        pkg.add_homepages(line.split(':', 1)[1])
-
-                    if line.startswith('# Maintainer:'):
-                        maintainer = line.split(':', 1)[1].strip()
-                        if ',' in maintainer:
-                            _, email = maintainer.split(',', 1)
-                            pkg.add_maintainers(extract_maintainers(email))
-                        else:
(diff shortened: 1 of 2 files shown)


TimB87:
I tested these changes today and they appear to be working perfectly fine. :+1:

AMDmi3:
Format and code look great. A few issues though:
- How does CRUX handle/support/EoL its releases? If e.g. 3.4, 3.5 are supported in parallel it would make sense to generate JSON and provide Repology entry for each release. For instance, Debian case shows that having an entry for abstract latest release while there are in fact versioned releases which people need to reference, is a bad idea
- `packagelinks` are gone, it's a regression. We could provide links to the latest (e.g. 3.5) Pkgfiles, but they will come out of sync with the new release. So, either links to some "latest" Pkgfiles are needed, or entries per release.
- Would it be too hard to support downloads (`source`) right away?
- One minor diff in data is url change `http://icedtea.classpath.org http://openjdk.java.net` ⇒ `http://icedtea.classpath.org` for contrib/openjdk8.

mhoush:
Thanks for the update and review.

* The only officially supported release is the current one (which is 3.5 now) - we will provide security updates for a previous release for a short time after the new release lands but we encourage all users to upgrade soon. This also means that the portdb and the repology JSON will always refer to the current supported release.
* I may misunderstand the point of `packagelinks` (as mentioned I'm new to repology) but my guess was that it's for providing links to built/finished packages for a distro. If that's correct, we don't provide those for CRUX, once you've installed from the install media, the rest is source-based. If I've misunderstood the point of `packagelinks` let me know and I can review further.
* For source downloads it would take some work to implement in the portdb; CRUX's `Pkgfile`s are bash but the portdb is python, so a shim would be needed to cross the gap. If this is a requirement I'll see what I can do but it doesn't currently exist.
* @TimB87 has volunteered to contact the maintainer of contrib/openjdk8.

AMDmi3:
- Let's leave an entry per release. It's much more convenient to provide consistent information tied to individual releases, than a virtual latest one, and leave info for older releases available since someone will stay on them even after EoL for sure.
- `packagelinks` primarily point to stuff related to source packages, e.g. build recipes. Pkgbuilds and their containing directories in case of CRUX.
- Having downloads available is not a requirement as long as homepage urls are supported, but it would be really useful for Repology (for matching packages), its users (more known download locations) and Repology would be able to report broken links in these. If you're not using bash to parse bash files, you'll have the same crippled parser as Repology currently has, which will break as soon as interpolation is encountered in name/version variables

mhoush:
* I don't mind leaving the old release entries, though this does present a problem in that there's no JSON dump available currently for those old releases, only the one to which the portdb points currently. That will not be hard to generate, though, so I can work on that. Should I just update this PR with a new version that brings those back but uses the JSON parser?
* I can add `packagelinks` pointing to the Pkgfiles for these repos, shouldn't be a problem.
* I think bash parsing is probably the easiest way to accomplish this. I looked at some things like bashlex to try to do it in python but the effort is probably not worth the gain. Homepage URLs are already there but I'll need to write and test the bash parsing and pass to python for source downloads.

AMDmi3:
- I don't mind removing old releases now as long as when 3.6 is released it would be possible to preserve 3.5, e.g. 3.5 and 3.6 would have separate json dumps. But the guarantee for this would basically be having per-release jsons now, and if these are available, there would be no need to remove older releases :)
- What I've meant is that you can't reliably parse bash without executing it. E.g. I'd do something like calling `bash --posix <(cat Pkgfile; echo set)`

AMDmi3:
God it's perfect. Thank you!

mhoush:
Awesome! Thanks for the help. :)

Issue #1078, opened 2020-09-21
[new] add repository - gentoo-science


epsilon-0:
The Gentoo Science repository is located at - https://github.com/gentoo/sci

This is the official place for adding scientific packages to Gentoo.

PR #1079, opened 2020-09-21
add gentoo science overlay

epsilon-0:
Adds the science project from Gentoo - https://wiki.gentoo.org/wiki/Project:Science

Related - https://github.com/repology/repology-updater/issues/1078

### DIFF
--- a/repos.d/gentoo/overlays.yaml
+++ b/repos.d/gentoo/overlays.yaml
@@ -59,6 +59,35 @@
   tags: [ all, production, gentoo-overlays ]
 
 ###########################################################################
+# Science
+###########################################################################
+- name: gentoo_ovl_science
+  type: repository
+  desc: Gentoo Science overlay
+  family: gentoo
+  color: 'f7a8b8'
+  minpackages: 100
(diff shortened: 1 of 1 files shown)


AMDmi3:
Sorry, this is not acceptable without metadata.

epsilon-0:
@AMDmi3 
if metadata were added, would it be possible to add this?

epsilon-0:
OK, so this metadata generation is stopped because of our current outdated state of the science repository.

Would it be possible to keep this open while we are fixing the repository and finish generating the metadata?

epsilon-0:
@AMDmi3 
I'm a bit confused on what you meant by **metadata**, but I've updated the link with the URL which contains the `md5-cache`.

Is this fine?

Thanks!

AMDmi3:
Yes, this is good. Thanks! Will deploy after adding some rules as this repository introduces a handful of new project name clashes.

epsilon-0:
@AMDmi3 
Thanks a lot!!!
:heart: :heart: :heart: :heart:

PR #1091, opened 2020-10-26
openwrt: un-shadow snapshot

mayli:
It pretty useful to show the package status for snapshot builds.

### DIFF
--- a/repos.d/embedded/openwrt.yaml
+++ b/repos.d/embedded/openwrt.yaml
@@ -95,7 +95,6 @@
       parser: OpenWrtPackagesParser
       url: 'https://downloads.openwrt.org/snapshots/packages/x86_64/{source}/Packages.manifest'
       subrepo: '{source}'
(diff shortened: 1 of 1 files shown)


AMDmi3:
This change is no-op as openwrt snapshot repository is completely disabled anyway.

mayli:
Any reason the snapshot repo is disabled? I think it quite useful to scan over snapshot to figure out outdated packages. It's not quite useful for current 19.07 release since lots of packages have been already updated to a newer version in snapshot.

AMDmi3:
Yes - it doesn't behave sanely with packages disappearing and reappearing over time. We can't use that as it pollutes the feeds a lot.

AMDmi3:
Merged. As mentioned, enabling the repo is impossible due to its unstable nature. Could use an alternative, preferrably some dump on openwrt source package metadata, but I haven't found anything like that.

PR #1106, opened 2020-12-27
updating liguros repos

palica:
- adding new releases and deprecating old

Please merge
Thank you

### DIFF
--- a/repos.d/gentoo/liguros.yaml
+++ b/repos.d/gentoo/liguros.yaml
@@ -29,10 +29,10 @@
       url: 'https://gitlab.com/liguros/{source}/-/tree/{branch}'
   packagelinks:
     - desc: Git directory of package
-      url: 'https://gitlab.com/liguros/{source}/-/tree/{branch}/{srcname}'
+      url: 'https://gitlab.com/liguros/liguros-repo/-/tree/develop/{srcname}'
     - desc: View ebuild
-      url: 'https://gitlab.com/liguros/{source}/-/blob/{branch}/{srcname}/{srcname|basename}-{rawversion}.ebuild'
-  tags: [ all, development, liguros ]
+      url: 'https://gitlab.com/liguros/liguros-repo/-/blob/develop/{srcname}/{srcname|basename}-{rawversion}.ebuild'
+  tags: [ all, production, liguros ]
 
 - name: liguros_21_1
   type: repository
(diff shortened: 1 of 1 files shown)


repos.d/gentoo/liguros.yaml
```diff
@@ -1,40 +1,38 @@
...
     - name:
-        - core-kit
-        - python-kit
-        - python-modules-kit
-        - nokit
+        - liguros-repo
       fetcher: GitFetcher
       parser: GentooGitParser
-      require_md5cache_metadata: true
+      require_md5cache_metadata: false
```
AMDmi3, line 19: Repositories without metadata are not acceptable.

repos.d/gentoo/liguros.yaml
```diff
@@ -1,40 +1,38 @@
...
       url: https://liguros.gitlab.io/
     - desc: GitLab Organization
       url: https://gitlab.com/liguros/
+    - desc: Official LiGurOS ebuild repository
+      url: 'https://gitlab.com/liguros/{source}/-/tree/{branch}'
   packagelinks:
-    - desc: Git directory
-      url: 'https://gitlab.com/liguros/{subrepo}/-/tree/20.7-release/{srcname}'
+    - desc: Git directory of package
+      url: 'https://gitlab.com/liguros/{source}/-/tree/{branch}/{srcname}'
```
AMDmi3, line 32: You can't use `{source}` and `{branch}` here.

repos.d/gentoo/liguros.yaml
```diff
@@ -1,40 +1,38 @@
...
     - desc: View ebuild
-      url: 'https://gitlab.com/liguros/{subrepo}/-/blob/20.7-release/{srcname}/{srcname|basename}-{rawversion}.ebuild'
-  tags: [ all, production, liguros ]
+      url: 'https://gitlab.com/liguros/{source}/-/blob/{branch}/{srcname}/{srcname|basename}-{rawversion}.ebuild'
+  tags: [ all, development, liguros ]
```
AMDmi3, line 35: The tag is useless. It's either production so it's available in repology, or it shouldn't be added to the config.

AMDmi3 (changes requested):

repos.d/gentoo/liguros.yaml
palica: https://github.com/repology/repology-updater/blob/master/repos.d/gentoo/sabayon.yaml#L15-L16
https://github.com/repology/repology-updater/blob/master/repos.d/gentoo/calculate.yaml#L15-L16

both repos have that set to false.

Where is the problem then to disable that for liguros repo? Please explain.

palica:
Requested changes have been commited. Please review/merge. Thank you

repos.d/gentoo/liguros.yaml
AMDmi3: Other repositories are not related to this case. Metadata is required, change which drops it is a pessimization and won't be accepted.

AMDmi3:
Timeout

PR #1110, opened 2021-01-14
Rename blackbox to backbox

elboulangero:
I believe it was a typo. The distro is actually named backbox, not blackbox.

This commit renames the yaml file, and also updates all references from 'blackbox' to 'backbox'.


AMDmi3:
Lol, nice catch

PR #1114, opened 2021-01-26
support for alpine 3.13

volkc:
Adds support for alpine 3.13 which has been released beginning of the year.

https://alpinelinux.org/posts/Alpine-3.13.0-released.html


volkc:
AFAIk the CI error is not related to the changes.

PR #1158, opened 2021-06-12
Fix Fedora 34 URL

fdw:
Fedora 34 has been released, and the URL should reflect that.

### DIFF
--- a/repos.d/rpm/fedora.yaml
+++ b/repos.d/rpm/fedora.yaml
@@ -338,10 +338,10 @@
   color: '294172'
   minpackages: 18000
   sources:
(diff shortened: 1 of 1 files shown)


AMDmi3:
Could you please update source name 3 lines above for consistency? Shouldn't 55 be added as well or is there no repo for it yet?

fdw:
> Could you please update source name 3 lines above for consistency?

Sure 🙂

> Shouldn't 55 be added as well or is there no repo for it yet?

I'm not a Fedora user (I just noticed the error in Repology), but I couldn't find a URL for it. So I guess not?

PR #1162, opened 2021-06-28
Update unitedrpms.yaml

kuboosoft:
Source change: some packages are missed


AMDmi3:
Please use a meaningful commit message. Also, giganet redirects, it may make sense to use `https://master.dl.sourceforge.net/project/unitedrpms/$v/srpm/repodata/repomd.xml?viasf=1` urls right away.

AMDmi3:
Timeout

Issue #1166, opened 2021-07-15
KISS source file changes


dylanaraps:
KISS now supports variables in its sources files. 

It looks like this:

```
https://zlib.net/zlib-VERSION.tar.gz
```

Which becomes:

```
https://zlib.net/zlib-1.2.11.tar.gz
```

Supported variable names are (example version file: `1.2.3+stuff-ddaefd 1`): 

- `VERSION` the first field of the version file as-is (`1.2.3+stuff-ddaefd`). 
- `RELEASE` the second field of the version file as-is (`1`).
- `MAJOR` the first component of `VERSION` split on `.+-_` (`1`).
- `MINOR` the second component of `VERSION` split on `.+-_` (`2`).
- `PATCH` the third component of `VERSION` split on `.+-_` (`3`).
- `IDENT` the rest of the components (starting from the fourth) as a single string (`stuff-ddaefd`).
- `PKG` the name of the package (the basename of the file's directory).

This works by simple substitution (marker is replaced by value). 
All matches on all lines are operated on.

While this feature is in the package manager, the repositories do not yet make use of it.
That change can be seen here: https://github.com/kisslinux/repo/pull/286

AMDmi3:
So this is becoming less and less of KISS, as expected. I don't plan to introduce extra complexity to the parser and duplicate the logic. Please provide a machine readable dump of relevant package data.

dylanaraps:
That's one way of looking at it. I don't feel the same way. Feel free to drop KISS from Repology as I have no plans to provide this information in a "fully compiled" form. No hard feelings whatsoever, thank you for supporting KISS thus far - I appreciate it.

illiliti:
I'm very against these new changes. I was in love with KISS because repository layout was so simple and without all of these nonsense variables. Now I feel betrayed... Therefore, I have to quit the development.

jedahan:
@AMDmi3 I am currently generating csv files for both the core kiss repository, and separately for all repositories. They are up on `http://jedahan.com/kiss-find/core.csv` and `http://jedahan.com/kiss-find/db.csv`. The format is roughly `name, version, repository_url, relative_path_to_package, optional_description`. Is there some documentation I can read to make it the most useful for repology, and get this upstreamed somehow?

AMDmi3:
[...]

PR #1169, opened 2021-07-22
Add noir

mmatongo:


AMDmi3:
We can't support this due to the same reason we no longer support kiss, see #1166.

AMDmi3:
We can't support this due to the same reason we no longer support kiss, see #1166.

mmatongo:
> We can't support this due to the same reason we no longer support kiss, see #1166.

I noticed that but we won't be using the new kiss variables for source files, instead we will maintain the old structure.

Will this have any affect on support?

AMDmi3:
If you don't use placeholders it should be OK. You currently do use them: https://github.com/noirlinux/main/blob/master/core/zlib/sources

mmatongo:
> If you don't use placeholders it should be OK. You currently do use them: https://github.com/noirlinux/main/blob/master/core/zlib/sources

They will be removed within the next couple of hours.

mmatongo:
I'm working on it right now.

AMDmi3:
Note that I don't in fact want to enforce upstream repository format in any way, I'm just not going to implement complex parsing. Alternatively, you could leave them (I see these as pretty convenient) be and provide a preprocessed package information dump for repology.

mmatongo:
I can go either way, but in this case to avoid being impeded I will most likely revert to the old format to avoid unnecessary complexities.

repos.d/kiss/noir.yaml
```diff
@@ -0,0 +1,33 @@
+###########################################################################
+# NOIR Linux
+###########################################################################
+- name: noir
+  type: repository
+  desc: NOIR Linux Main
+  family: noir
```
AMDmi3, line 7: You also need to list this in [repology-schemacheck.py](repology-schemacheck.py)

mmatongo: done.

mmatongo:
> If you don't use placeholders it should be OK. You currently do use them: https://github.com/noirlinux/main/blob/master/core/zlib/sources

This has been cleared up now.

PR #1179, opened 2021-09-28
Add NUR

wamserma:
[NUR](https://nur.nix-community.org/) is to the Nixpkgs/NixOS ecosystem what AUR is for Arch.
The main difference is that packages are grouped by maintainers, which gives some additional flexibility to users (choice between full NUR or only some selected maintainers, having different package flavours available without namespace cluttering).


AMDmi3 (changes requested):

repos.d/nix/nur.yaml
```diff
@@ -0,0 +1,34 @@
...
+  type: repository
+  desc: NUR combined
+  statsgroup: nix
+  family: nix
+  ruleset: [nix]
+  color: '7eb2dd'
+  update_period: 24h
+  minpackages: 100
+  default_maintainer: fallback-mnt-aur@repology
+
```
AMDmi3, line 15: no need for empty lines

repos.d/nix/nur.yaml
```diff
@@ -0,0 +1,34 @@
...
+###########################################################################
+- name: nur
+  sortname: nix-unstable~nur  # place right after Nix unstable
+  type: repository
+  desc: NUR combined
+  statsgroup: nix
+  family: nix
+  ruleset: [nix]
+  color: '7eb2dd'
+  update_period: 24h
```
AMDmi3, line 12: We can update it more often (e.g. remove this line). AUR is a problem because there's no single file with all the info and we have to paginate through the API, this is likely not the cause here.

AMDmi3 (changes requested):

repos.d/nix/nur.yaml
wamserma: AFAIK the info is regenerated once a day, but I agree: Let's keep artificial delays low.

repos.d/nix/nur.yaml
```diff
@@ -0,0 +1,29 @@
...
+      parser:
+        class: NixJsonParser
+  repolinks:
+    - desc: NUR home
+      url: https://nur.nix-community.org/
+    - desc: GitHub repository
+      url: https://github.com/nix-community/NUR
+  packagelinks:
+    - type: PACKAGE_RECIPE
+      url: 'https://github.com/nix-community/nur-combined/blob/master/{posfile}#L{posline}'
```
AMDmi3, line 28: This produces broken URLs. The dump contains ready to use URLs but they are broken (trimmed up to repository path) sometimes as well.

AMDmi3 (changes requested):
This does not work as the dump format is different from one of nix dump. At the very least, the latter contains package info under top level "packages" dict item. So this may need a custom parser, or nix parser parametrized to support different format.

repos.d/nix/nur.yaml
wamserma: deactivated that part, like it's done for Chaotic-AUR

repos.d/nix/nur.yaml
```diff
@@ -0,0 +1,29 @@
+###########################################################################
+# NUR
+###########################################################################
+- name: nur
+  sortname: nix-unstable~nur  # place right after Nix unstable
```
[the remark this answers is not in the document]
wamserma: fixed

wamserma:
> This does not work as the dump format is different from one of nix dump. At the very least, the latter contains package info under top level "packages" dict item. So this may need a custom parser, or nix parser parametrized to support different format.

I will look into that.

repos.d/nix/nur.yaml
wamserma: Upon further investigation all position-URLs in the dump look fine to me.

AMDmi3: > Upon further investigation all position-URLs in the dump look fine to me.

`curl -s https://raw.githubusercontent.com/nix-community/nur-search/master/data/packages.json | grep \"position\" | grep -v \#`

wamserma: Ah, I misunderstood your first comment. These are really tricky ones, e.g. for `nur.repos.arc.packages.imv-develop` the JSON points to `https://github.com/nix-community/nur-combined/tree/master/repos/arc` but should actually point to [https://github.com/nix-community/nur-combined/blob/master/repos/arc/pkgs/customized/default.nix#L92](https://github.com/nix-community/nur-combined/blob/b48688290f4168bc1980c99caa2220f383ea90a5/repos/arc/pkgs/customized/default.nix#L92), which is a function that transforms the upstream package from nixpkgs and includes the result in NUR.
I think these packages should be filtered out.

PR #1181, opened 2021-09-29
Fedora 35

keszybz:


repos.d/rpm/fedora.yaml
```diff
@@ -386,6 +386,47 @@
...
+    - desc: Fedora repository for package maintenance
+      url: https://src.fedoraproject.org/
+  packagelinks:
+    - type: PACKAGE_SOURCES
+      url: 'https://src.fedoraproject.org/rpms/{srcname}/tree/f35'
+    - type: PACKAGE_RECIPE
+      url: 'https://src.fedoraproject.org/rpms/{srcname}/blob/f35/f/{srcname}.spec'
+    - type: PACKAGE_RECIPE_RAW
+      url: 'https://src.fedoraproject.org/rpms/{srcname}/raw/f35/f/{srcname}.spec'
+  tags: [ all, fedora, rpm ]
```
AMDmi3, line 428: Please add `production` tag here for it to appear on main Repology instance

repos.d/rpm/fedora.yaml
```diff
@@ -386,6 +386,47 @@
...
   tags: [ all, production, fedora, rpm ]
 
+- name: fedora_35
+  type: repository
+  desc: Fedora 35
+  statsgroup: Fedora
+  family: fedora
+  ruleset: [fedora, rpm]
+  color: '294172'
+  minpackages: 25000
```
AMDmi3, line 396: This fails as it only has 22581 packages. You should probably leave 22000 everywhere, is it's the fair number of packages without duplicates.

AMDmi3 (changes requested):

repos.d/rpm/fedora.yaml
keszybz: Oh, this is about source packages… Yeah, I think we have 23.5k now.

AMDmi3 (approved):

AMDmi3:
Thanks you!

Issue #1191, opened 2021-11-03
AppGet is dead


algent-al:
TLDR from Defeloper: 
> I’m no longer going to be developing AppGet. The client and backend services will go into maintenance mode immediately until August 1st, 2020, at which point they’ll be shut down permanently.

Read more here: https://keivan.io/the-day-appget-died/

AMDmi3:
I see no reason to remove it still.

PR #1194, opened 2021-11-12
Remove appget as it’s shut down

rasa:
See https://github.com/appget/appget#appget-is-shutting-down


AMDmi3:
#1191

rasa:
I searched, but didn't find that issue, sorry.

AMDmi3:
The change itself is good though, I might merge it if it's decided to remove appget.

PR #1201, opened 2021-11-26
NixOS: Update to 21.11.

heroku-miraheze:
You will merge these changes until NixOS creates nixos-21.11.

### DIFF
--- a/repos.d/nixos.yaml
+++ b/repos.d/nixos.yaml
@@ -1,7 +1,7 @@
 ###########################################################################
 # NixOS packages
 ###########################################################################
-- name: nix_stable
+- name: nix_stable_21_11
   type: repository
   desc: nixpkgs stable 21.11
   statsgroup: nix
--- a/repos.d/nixos.yaml
+++ b/repos.d/nixos.yaml
@@ -1,12 +1,12 @@
 ###########################################################################
 # NixOS packages
 ###########################################################################
-- name: nix_stable  # XXX: name the next one as e.g. nix_stable_21_11
+- name: nix_stable_21_11
   type: repository
   desc: nixpkgs stable 21.11
   statsgroup: nix
   family: nix
-  ruleset: [nix,nix_old_node_naming]  # don't set nix_old_node_naming for next release
+  ruleset: [nix]  # don't set nix_old_node_naming for next release
   color: '7eb2dd'
   minpackages: 70000
   default_maintainer: fallback-mnt-nix@repology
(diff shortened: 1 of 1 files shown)


AMDmi3:
There's one more line in `packagelinks` which needs to be updated.

AMDmi3:
Also there are two comments about what needs to be done on update.

heroku-miraheze:
We already created NixOS 21.11, so, you can merge it now.

AMDmi3:
Please include remaining requested changes.

heroku-miraheze:
Updated.

heroku-miraheze:
Please, merge it now.

AMDmi3:
The comments say exactly what should be done. Also please squash into a single commit.

tomberek:
We're ready to go. What's the best way to do this for the next release? Is there a process that work best here?

repos.d/nixos.yaml
```diff
@@ -1,20 +1,20 @@
 ###########################################################################
 # NixOS packages
 ###########################################################################
-- name: nix_stable  # XXX: name the next one as e.g. nix_stable_21_05
+- name: nix_stable  # XXX: name the next one as e.g. nix_stable_21_11
```
vcunat, line 4:
```suggestion
- name: nix_stable_21_11
```

vcunat:
For the record:
- 21.11 was officially released several hours ago, e.g. https://nixos.org/blog/announcements.html#21.11
- 21.05 still has promised support this December, but I think it's OK to drop it from repology in this commit already (as it does now)

vcunat (approved):

repos.d/nixos.yaml
```diff
@@ -1,20 +1,20 @@
...
   type: repository
-  desc: nixpkgs stable 21.05
+  desc: nixpkgs stable 21.11
   statsgroup: nix
   family: nix
-  ruleset: [nix,nix_old_node_naming]  # don't set nix_old_node_naming for next release
+  ruleset: [nix]  # don't set nix_old_node_naming for next release
   color: '7eb2dd'
-  minpackages: 60000
+  minpackages: 70000
```
vcunat, line 11: Wait, are you sure?  For unstable it shows less ATM: https://repology.org/repositories/statistics/total

vcunat: On the whole I can't see motivation to try driving this very close to the real number anyway, but maybe there is some.

AMDmi3: >  For unstable it shows less ATM in the "total" column

This column reports a number of projects (aggregate entities), not individual packages. See [packages](https://repology.org/repositories/packages) tab or [parse logs](https://repology.org/log/9555990) in `updates` tab instead.

> On the whole I can't see motivation to try driving this very close to the real number anyway, but maybe there is some.

The purpose for this knob is to catch cases where upstream suddenly generates broken incomplete dump (and there were such cases on practice) so it makes sense to keep it close. It's not very important though.

AMDmi3:
It's almost good, I'll add finishing touches myself, squash and merge.

> What's the best way to do this for the next release? Is there a process that work best here?

The following updates would only require cloning the latest entry and updating the version in 4 lines.
Older entries should be preserved for at least their support period, and may in fact be preserved forever (or as long as the [channel](https://channels.nixos.org/) is available). Filling `valid_till: YYYY-MM-DD` is preferrable in the latter case for uses like repology/repology-webapp#117.

> 21.05 still promises support this December, but I think it's OK to drop it from repology in this commit already (as it does now)

I think it makes sense to preserve 21.05 entry for some time in this case. If it EoLs in Jan, I'd remove it then in order to also remove related legacy hacks. Following branches may be preserved for longer time.

vcunat:
The promised support period is always in release announcement (and always 7 months so far), so `valid_till` can be used.

AMDmi3:
@heroku-miraheze changes merged as 288ccea6fa1e1f6081b8dba28cf47cba6a0df5ae.

PR #1206, opened 2021-12-15
rpmfusion: add 34/35, remove 31 32

kwizart:
This adds the 34 and 35 versions of the repository and drop the oudated 31 and 32 ones (like Fedora).

Signed-off-by: Nicolas Chauvet <kwizart@gmail.com>


kwizart:
Forced pushed a new version of the commit to not remove older releases.

Also added tainted free/nonfree to better advertise theses content.

PR #1207, opened 2021-12-17
Add openSUSE OBS "shells" repo for Tumbleweed

milliams:
This contains, e.g. nushell (https://software.opensuse.org/package/nushell) which use repology to list their packages (https://github.com/nushell/nushell#packaging-status).


AMDmi3:
Merged, thanks!

PR #1216, opened 2022-01-13
NixOS 21.05 is EOL, remove

jonringer:
No longer officially supported.

https://nixos.org/manual/nixos/stable/release-notes.html#sec-release-21.05

### DIFF
--- a/repos.d/nixos.yaml
+++ b/repos.d/nixos.yaml
@@ -1,35 +1,6 @@
 ###########################################################################
 # NixOS packages
 ###########################################################################
-- name: nix_stable  # XXX: remove in 2022
-  type: repository
-  desc: nixpkgs stable 21.05
-  statsgroup: nix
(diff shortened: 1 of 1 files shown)


AMDmi3:
No need to as long as it's available

jonringer:
> No need to as long as it's still available

It will always be available. The endpoints aren't removed: https://channels.nixos.org/

AMDmi3:
Then it can live in Repology for indefinite amount of time. Removing it only saves some resources and shortens some tables on the website, while leaving it provides useful historical data.

vcunat:
@AMDmi3: but 21.05 is still shown here as the default `nixos_stable`, and that's clearly wrong.  So we should do some kind of renames or something?

AMDmi3:
It's not default, it's just an unique identifier.

Issue #1222, opened 2022-02-06
Feb 2022: a bunch of broken repositories


AMDmi3:
During the last few days, a lot of repositories (independently) broke in the parse phase. This is critical because repositories failing this way prevent applying new rules. There's no way to fix it gracefully short-term other than removal of such repositories, and a long-term fix would require a massive rewrite of the update process and introduce a dependency on e.g. ZFS snapshots (to introduce a possibility to roll back to a last good state for any repository).

Failing repos:
- [x] AUR: suddenly stopped providing required `URL` field after changes in aurweb. Probably because of [this](https://github.com/archlinux/aurweb/commit/29c2d0de6b83a2287ffdb885d9b55aa63b1d4792#r66116492) change (ping @kevr), also affects MPR (#1161) which was removed some weeks ago. 
- [x] FreeBSD: a stray port without version defined. Already reported and [fixed](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=261745)
- [x] Funtoo: [broken XML](https://github.com/funtoo/desktop-kit/blob/fb358848968b47b3d30c6272c220d991ce2f0e4b/x11-misc/i3lock/metadata.xml) for x11-misc/i3lock in desktop-kit. Ping @palica, @danielrobbins? 
- [ ] Nix (unstable): suddenly stopped providing most of package data (`meta` dict for each package). Ping @edolstra?
- [x] Slackbuilds: missing quote https://git.slackbuilds.org/slackbuilds/commit/multimedia/minitube/minitube.info?id=457c0ba4f2b63ebedbe748d19056a123f06a7649. Ping @willysr.
- [x] Termux: broken json - already [fixed](https://github.com/termux/termux-packages/commit/b760237d8831071490ed9585dc7a8083c940ca2c) (see also comments)

Since these failures prevent fixing a bunch of problem reports which accumulated during past two weeks of my vacation, I can't wait too long for these to be fixed, so let's say there's time till Wednesday for these repositories to fix themselves, otherwise I'll have to remove them until the problems are resolved.

Also a couple of sources are failing in fetch phase now, which is not that critical, but is never the less a good reason to get rid of these too.
- [x] distrowatch (HTTP 400, they have mistakenly banned Repology before and now it happened again) fixed
- [ ] hyperbola (repository data disappeared)

willysr:
Fixed in rebuild-fixes
[...]

PR #1225, opened 2022-02-09
Reduce nix perl package noise

jonringer:
The nix bump from 2.5.1 to 2.6 exposed a bug in nix around retaining position information with the usage of `builtins.catAttrs`.

Until there's a proper fix, just return semi-accurate information.

related: https://github.com/repology/repology-updater/issues/1222 https://github.com/NixOS/nix/issues/6068 https://github.com/NixOS/nixpkgs/pull/158548

### DIFF
--- a/repology/parsers/parsers/nix.py
+++ b/repology/parsers/parsers/nix.py
@@ -211,6 +211,15 @@ class NixJsonParser(Parser):
                         # XXX: uses MELPA versions which are timestamps
                         pkg.set_flags(PackageFlags.UNTRUSTED)
                 else:
+                    # perl packages have improper file position due to broken builtin operation
+                    # https://github.com/NixOS/nix/issues/6068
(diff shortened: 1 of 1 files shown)


jonringer:
I would prefer to attempt to fix this in nixpkgs first.

jonringer:
nevermind, this is polling the `nixos-unstable` [endpoint](https://github.com/repology/repology-updater/blob/18f353331def303681193e5e9fd9199506bb96f0/repos.d/nixos.yaml#L75). 24hrs should see the fix propagate

AMDmi3:
There is another problem now:

https://repology.org/project/obsidian/versions

dwarf fortress package which has no prefix and an ambiguous name is clashing with unrelated project.

jonringer:
Since this is a PR, I'm going to move the nixpkgs issues to https://github.com/repology/repology-updater/issues/1226

PR #1232, opened 2022-02-26
Add CentOS Stream 9

AlexBaranowski:
Hi,

I want to propose these changes to add CentOS Stream 9.

When it comes to GitLab links, I tried with a few packages, and it worked well.

Best,
Alex

### DIFF
--- a/repos.d/rpm/centos.yaml
+++ b/repos.d/rpm/centos.yaml
@@ -89,9 +89,9 @@
       url: 'https://git.centos.org/rpms/{srcname}/raw/c8{centossuffix}/f/SPECS/{srcname}.spec'
   groups: [ all, production, centos, rpm ]
 
-- name: centos_8_stream
+- name: centos_stream_8
   type: repository
-  desc: CentOS 8 Stream
+  desc: CentOS Stream 8
   statsgroup: CentOS
   family: centos
   ruleset: [centos, rpm]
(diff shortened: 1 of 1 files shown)


repos.d/rpm/centos.yaml
```diff
@@ -119,6 +119,35 @@
...
+  repolinks:
+    - desc: CentOS home
+      url: https://www.centos.org/
+  packagelinks:
+    - type: PACKAGE_SOURCES
+      url: 'https://gitlab.com/redhat/centos-stream/rpms/{srcname}/-/tree/c9s'
+    - type: PACKAGE_RECIPE
+      url: 'https://gitlab.com/redhat/centos-stream/rpms/{srcname}/-/blob/c9s/{srcname}.spec'
+    - type: PACKAGE_RECIPE_RAW
+      url: 'https://gitlab.com/redhat/centos-stream/rpms/{srcname}/-/blob/c9s/{srcname}.spec'
```
Conan-Kudo, line 148: This should be changed like so:

```suggestion
      url: 'https://gitlab.com/redhat/centos-stream/rpms/{srcname}/-/raw/c9s/{srcname}.spec'
```

Conan-Kudo (changes requested):

repos.d/rpm/centos.yaml
AlexBaranowski: @Conan-Kudo  You are right! I missed it. Nice catch.

AMDmi3:
Is it centos stream 9 or centos 9 stream? Cause if the former is the case we probably need to rename centos 8 stream.

Conan-Kudo:
The official name is "CentOS Stream 9". The older major version is "CentOS Stream 8".

AlexBaranowski:
@AMDmi3 @Conan-Kudo I changed repo name and description for CentOS Stream 8

AMDmi3:
Merged, thanks!

PR #1233, opened 2022-02-26
Add EuroLinux 8

AlexBaranowski:
Hi,

I would love to add EuroLinux to Your project. I adapted the Scientific Linux config, and I decided to put it into a separate file, as CentOS 8 is dead and CentOS Stream 8 has a different purpose.

Best,
Alex


AMDmi3:
Merged, thanks!

PR #1238, opened 2022-03-17
Fix link to nixpkgs-unstable

mausch:
I could be wrong but think this should point to the nixpkgs-unstable branch not master?

@jonringer

### DIFF
--- a/repos.d/nixos.yaml
+++ b/repos.d/nixos.yaml
@@ -84,5 +84,5 @@
       url: https://github.com/NixOS/nixpkgs
   packagelinks:
     - type: PACKAGE_RECIPE
-      url: 'https://github.com/NixOS/nixpkgs/blob/master/{?posfile}#L{?posline}'
+      url: 'https://github.com/NixOS/nixpkgs/blob/nixpkgs-unstable/{?posfile}#L{?posline}'
   groups: [ all, production, have_testdata, nix ]
--- a/repos.d/nixos.yaml
+++ b/repos.d/nixos.yaml
@@ -84,5 +84,5 @@
       url: https://github.com/NixOS/nixpkgs
   packagelinks:
     - type: PACKAGE_RECIPE
-      url: 'https://github.com/NixOS/nixpkgs/blob/nixpkgs-unstable/{?posfile}#L{?posline}'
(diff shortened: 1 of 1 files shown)


AMDmi3:
What's the difference between these branches then?

AMDmi3:
> See also https://nixos.wiki/wiki/Nix_channels

This one says:

> Unstable channels (nixos-unstable, nixpkgs-unstable) correspond to the main development branch (master)

jonringer:
on line 75, you can see that it points to `nixos-unstable`, which would be better. However, I don't feel strongly either way. There's always going to be the potential for these things to be misaligned.

> What's the difference between these branches then?

`master` is the development branch. the `nixos-*` and `nixpkgs-*` branches are "release branches", they get incremented by CI/CD machinery when certain gates pass, and all builds are attempted.

I would advise against using the `nixpkgs-*` channels, as they also include less supported platforms like MacOS, and this can cause the metadata to be significantly more stale.

> More importantly here the sources come from https://channels.nixos.org/nixos-unstable/packages.json.br , there isn't a similar URL for master AFAIK.

The top-level contents of a change (e.g. [this](https://releases.nixos.org/nixos/unstable/nixos-22.05pre360796.73ad5f9e147)) are only generated by the CI/CD machinery, they do not really "exist" on the master branch. (although it can be created by doing `$ nix-build ./nixos/release.nix -A channel`)

repos.d/nixos.yaml
```diff
@@ -84,5 +84,5 @@
       url: https://github.com/NixOS/nixpkgs
   packagelinks:
     - type: PACKAGE_RECIPE
-      url: 'https://github.com/NixOS/nixpkgs/blob/master/{?posfile}#L{?posline}'
+      url: 'https://github.com/NixOS/nixpkgs/blob/nixpkgs-unstable/{?posfile}#L{?posline}'
```
jonringer, line 87: Use same release channel as the metadata updater
```suggestion
      url: 'https://github.com/NixOS/nixpkgs/blob/nixos-unstable/{?posfile}#L{?posline}'
```

AMDmi3:
Merged, thanks!

PR #1242, opened 2022-04-08
Remove NixOS 21.05

ryantm:
It is no longer supported. I wasn't sure if I'm supposed to rename the `nix_stable_21_11` to `nix_stable`.


AMDmi3:
We don't remove repositories which are still available

PR #1255, opened 2022-06-23
Add APK Fission for Adélie repo

VulpineAmethyst:
APK Fission has been part of the Adélie ecosystem for several years now, but until recently was not able to fully support all of Adélie's Tier 1 architectures. This repository does not contain any PPAs, only software that Adélie does not ship (including non-free software).

### DIFF
--- a/repos.d/apkfission.yaml
+++ b/repos.d/apkfission.yaml
@@ -1,0 +1,42 @@
+###########################################################################
+# APK Fission packages
+#
+# https://apkfission.net/install.html
+###########################################################################
+
+{% macro apkfission(version, archs=['aarch64','armv7','pmmx','ppc','ppc64','x86_64']) %}
+- name: apkfission_adelie_{{version|replace('.', '_')}}
+  type: repository
+  desc: APK Fission for Adélie {{version|title}}
+  statsgroup: APKFission
+  family: apkfission
+  ruleset: adelie
+  default_maintainer: fallback-mnt-apkfission@repology
+  sources:
+    {% for subrepo in ['main', 'nonfree'] %}
+    - name: {{subrepo}}/x86_64
+      fetcher:
+        class: TarFetcher
+        url: 'http://distfiles.apkfission.net/adelie/{{version}}/{{subrepo}}/x86_64/APKINDEX.tar.gz'
+      parser:
+        class: ApkIndexParser
+      subrepo: {{subrepo}}/x86_64
+    {% endfor %}
+  repolinks:
+    - desc: APK Fission home
+      url: https://apkfission.net/
+  minpackages: 50
+  packagelinks:
+    {% for subrepo in ['main', 'nonfree'] %}
+    - type: PACKAGE_SOURCES
+      url: 'https://git.adelielinux.org/apkfission/packages/-/tree/master/{{subrepo}}/{srcname}'
+    - type: PACKAGE_RECIPE
+      url: 'https://git.adelielinux.org/apkfission/packages/-/blob/master/{{subrepo}}/{srcname}/APKBUILD'
+    - type: PACKAGE_RECIPE_RAW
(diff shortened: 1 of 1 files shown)


repos.d/apkfission.yaml
```diff
@@ -0,0 +1,40 @@
+###########################################################################
+# APK Fission packages
+#
+# https://apkfission.net/install.html
+###########################################################################
+
+{% macro apkfission(version, subrepos=['main','nonfree'], archs=['aarch64','armv7','pmmx','ppc','ppc64','x86_64']) %}
```
AMDmi3, line 7: This probably belongs to adelie.yaml instead a dedicated file

repos.d/apkfission.yaml
```diff
@@ -0,0 +1,40 @@
...
+- name: apkfission_adelie_{{version|replace('.', '_')}}
+  type: repository
+  desc: APK Fission for Adélie {{version|title}}
+  statsgroup: APKFission
+  family: apkfission
```
AMDmi3, line 12: I don't think this needs a dedicated family (since it's realted to adelie), also it needs a `ruleset` (to apply the same rules as for adelie). May go with the same values as adelie.

repos.d/apkfission.yaml
AMDmi3, line 7: If subrepos are unlikely to change, I'd move these out of parameters.

AMDmi3, line 7: Webapp does not currently handle a lot of similar packages, e.g. a package of the same thing for a lot of arches. That's why we only stich to ppc64 for adelie currently. I'd prefer the same here for now.

repos.d/apkfission.yaml
```diff
@@ -0,0 +1,40 @@
...
+        class: ApkIndexParser
+      subrepo: {{subrepo}}/{{arch}}
+    {% endfor %}
+    {% endfor %}
+  repolinks:
+    - desc: APK Fission home
+      url: https://apkfission.net/
+  packagelinks:
+    - type: PACKAGE_SOURCES
+      url: 'https://git.adelielinux.org/apkfission/packages/-/tree/master/{subrepo}/{srcname}'
```
AMDmi3, line 31: This may be confusing, but it takes subrepo set in `sources` to `{{subrepo}}/{{arch}}`, so it'll include arch which you don't want. You may move `packagelinks` block to `source` and use Jinja templates there, e.g. 'https://git.adelielinux.org/apkfission/packages/-/tree/master/{{subrepo}}/{srcname}'

repos.d/apkfission.yaml
AMDmi3, line 12: This also needs `minpackages`

AMDmi3 (changes requested):

repos.d/apkfission.yaml
VulpineAmethyst: Unfortunately, APK Fission's packaging has numerous differences between archs (firmware specific to aarch64, microcode specific to pmmx/x86_64, etc.), so I do need to have arch-related subrepos. I could pin to x86_64 for now since that's necessary, but I'd really like to not have to.

repos.d/apkfission.yaml
VulpineAmethyst: The plan is to eventually start packaging for Alpine too; I just wanted to get all of my ducks in a row with Adélie support first.

repos.d/apkfission.yaml
awilfox: APK Fission is similar to, and named somewhat after, RPM Fusion – which has its own file despite being only for Fedora.  It has no organisational connection to Adélie.

repos.d/apkfission.yaml
awilfox: I agree that the Adélie repo should have the `adelie` `ruleset`; the Alpine repo can have the `alpine` one.

VulpineAmethyst:
This is no longer necessary, as the APK Fission project has been abandoned.

PR #1259, opened 2022-07-03
Update opensuse.yaml with correct update repos

mimi1vx:
openSUSE from Leap 15.3 changed structure of update repos with full binary compatibility with SLE + removed out of support distributions

### DIFF
--- a/repos.d/rpm/opensuse/opensuse.yaml
+++ b/repos.d/rpm/opensuse/opensuse.yaml
@@ -263,6 +263,20 @@
       parser:
         class: RepodataParser
       subrepo: updates/non-oss
+    - name: backports-update
+      fetcher:
+        class: RepodataFetcher
+        url: http://download.opensuse.org/update/leap/15.3/backports/
+      parser:
+        class: RepodataParser
+      subrepo: updates/backports
+    - name: sle-update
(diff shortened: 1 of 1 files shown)


AMDmi3:
Repository removals will not be accepted

mimi1vx:
@AMDmi3 why? Leap 42.3 + Leap 15.0-15.2 are dead and only polute results and page with obsolete data. And in the end gives wrong impressions about totally outdated openSUSE ...

AMDmi3:
For history and for these who still use these repositories.

mimi1vx:
> For history and for these who still use these repositories.

history ok, this is valid, usage of this repos is, how to say, total stupidity?

mimi1vx:
@AMDmi3 now, with only with fixes for Leap 15.3/Leap 15.4

mimi1vx:
@AMDmi3  .. question is simple .. distribution is based on this. So its enabled by default.

AMDmi3:
Sorry, I've forgot about this and it has rot somewhat. Remaining relevant part was about adding backports and sle subrepos, and I've added these in c68343d74363f278c0afb56c7af70d2ba4039b2d

PR #1265, opened 2022-08-07
add gentoo wayland overlay

epsilon-0:
Signed-off-by: Aisha Tammy <aisha@bsd.ac>


AMDmi3:
Sorry, we don't add overlays

PR #1271, opened 2022-09-23
Add openEuler 22.03

robert-scheck:
See also: https://docs.openeuler.org/en/docs/22.03_LTS/docs/Releasenotes/release_notes.html


AMDmi3:
merged in b3c6aded7aa3f49ee60c669e3b7d752dd3869219

Issue #1270, opened 2022-09-22
Merge EuroLinux into centos.yaml?


robert-scheck:
[EuroLinux](https://en.euro-linux.com/eurolinux/what-is/) is from my point of view just another RHEL clone/derivative; citing their website:

> EuroLinux is a Polish Enterprise-class Linux distribution that has been in development since 2013. EuroLinux is built upon the foundation of Red Hat® Enterprise Linux® source code. This guarantees its compatibility with RHEL®, Oracle® Linux, CentOS, as well as new, recently published community distributions AlmaLinux and Rocky Linux. EuroLinux offers the same operating environment of software designed for these systems, and provides the same hardware certification.

Given Rocky Linux and AlmaLinux are also part of `centos.yaml`, I suggest to merge EuroLinux's separate yaml file into it. Or, alternatively, splitting Rocky Linux and AlmaLinux into separate yaml files could make sense, too.

PR #1272, opened 2022-09-23
Add EuroLinux 9

robert-scheck:
See also: https://en.euro-linux.com/blog/eurolinux-9-0-released/


AMDmi3:
This doesn't work as these repositories do not contain any source packages.

robert-scheck:
I'm sorry, corrected.

AMDmi3:
This loses `{subrepo}` in the link, but let me just convert `eurolinux_8` into template, reuse it for 9 and move it to centos config along the way (#1270).

PR #1275, opened 2022-09-24
Add "package page" for Fedora and EPEL

robert-scheck:


AMDmi3:
This does not seem completely correct, as it `https://packages.fedoraproject.org/pkgs/{srcname}/{srcname}` refers to a subpackage named the same way as the source package, and
- I'm not sure if subpackage with the same always exists
- It refers to a different entity than a whole source package
So it looks like it should be trimed to just `https://packages.fedoraproject.org/pkgs/{srcname}`

robert-scheck:
> I'm not sure if subpackage with the same always exists

Yes, there could be some corner cases where this might not apply, thus changed as proposed by you. Good catch!

AMDmi3:
I've missed one more thing, should it not apply to `epel_6` as well?

robert-scheck:
No, because EPEL 6 is end-of-life and thus EPEL 6 builds (or packages that only existed on that branch) won't show up at packages.fedoraproject.org.

PR #1277, opened 2022-09-27
Removed CRUX versions 3.4 and 3.5 as they're unsupported, undeveloped…

mhoush:
…, and unmaintained


AMDmi3:
We don't remove outdated repositories. But you could add EOL dates (`valid_till`) to them.

PR #1279, opened 2022-09-28
Added valid_till dates to old CRUX versions

mhoush:
OK, does this look like the correct way to do it instead of #1277?


AMDmi3:
Yes, it's perfect, thank you.

mhoush:
Thanks!

PR #1284, opened 2022-10-22
Packages that don't have urls don't have versions

firasuke:
By skipping packages that don't have `url`s we're also skipping packages that don't have a version number (a package that has a `url` must have a version number and a checksum, but the latter is not true, meaning that checking for `url` only should suffice to determine virtual packages).

This should simplify things by a bit.


AMDmi3:
I'd prefer to keep the check for logs verbosity and just for the case when something changes.

firasuke:
I see, that makes sense.

PR #1292, opened 2022-11-24
Add the debian janitor repositories

jelmer:
These are from the Debian Janitor project to automatically pull in new upstream releases and snapshots.

They're essentially overlay repositories, like e.g. Debian experimental, and do not contain the full set of packages but are intended to be used on top of e.g. unstable.

### DIFF
--- a/repos.d/deb/janitor.yaml
+++ b/repos.d/deb/janitor.yaml
@@ -1,0 +1,69 @@
+###########################################################################
+# Debian Janitor
+#
+# https://janitor.debian.net/fresh
+###########################################################################
+
+- name: debian_janitor_releases
+  type: repository
+  desc: Debian Janitor - Fresh Releases
+  statsgroup: Debian+derivs
+  family: debuntu
+  color: '32638f'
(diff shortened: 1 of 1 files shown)


AMDmi3:
https://janitor.debian.net/dists/fresh-releases/contrib/source/Sources.gz → 404
https://janitor.debian.net/dists/fresh-snapshots/contrib/source/Sources.gz → 404

jelmer:
> https://janitor.debian.net/dists/fresh-releases/contrib/source/Sources.gz → 404 https://janitor.debian.net/dists/fresh-snapshots/contrib/source/Sources.gz → 404

Fixed; the Janitor just publishes "main" as a compoment, so I've dropped contrib and non-free.

AMDmi3:
These lack `packagelinks` at least to package sources.

jelmer:
> These lack `packagelinks` at least to package sources.

Added.

jelmer:
It looks like this commit made its way in, so I'll close this PR. Thanks, @AMDmi3 !

PR #1293, opened 2022-11-24
Add the Kali bleeding-edge repository

jelmer:


jelmer:
@AMDmi3 rebased after your last commit to disable the flake8 step

Issue #1299, opened 2023-01-01
OpenBSD packages should link to openports.pl


morgant:
Currently, OpenBSD packages link to openports.se, but that has been known to be unreliable at times due to occasional/intermittent issues with scraping the OpenBSD ports repository. While not hosted at OpenBSD.org for security reasons, an OpenBSD developer maintains openports.pl and it is preferred over openports.se. That said, openports.pl doesn't retain some of the historical information that openports.se does, so I don't suggest replacing the latter, just adding the former (preferably more prominently).

morgant:
Yup, looks good! Thanks again, @AMDmi3!

PR #1300, opened 2023-01-01
Add openports.pl links for OpenBSD packages

morgant:
This adds links to openports.pl to OpenBSD packages. Links to openports.se are retained. See Issue #1299.

### DIFF
--- a/repos.d/bsd/openbsd.yaml
+++ b/repos.d/bsd/openbsd.yaml
@@ -23,11 +23,15 @@
   repolinks:
     - desc: 'OpenBSD FAQ: Packages and Ports'
       url: https://www.openbsd.org/faq/faq15.html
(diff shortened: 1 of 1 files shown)


morgant:
@AMDmi3, my apologies for pinging you directly, but I'm hoping this could be reviewed and potentially merged.

AMDmi3:
On the coincidence, I was working on this just now. The change is good, I'm merging it.

The problem (which has caused the delay) though is that link preference (according to the order in repository config) is not respected right now, so pages which have to pick one link (such as `project/FOO/versions`) may still use openports.se links. The fix for that is ready, I'll deploy it tomorrow after some testing.

morgant:
@AMDmi3 Ah, many, many thanks!

PR #1304, opened 2023-02-01
nixpkgs: extract GitHub maintainers

ncfavier:
A recent change (https://github.com/NixOS/nixpkgs/pull/209165) removed `noreply.github.com` email addresses from the maintainer list, which means that maintainers using only a GitHub account are not indexed by repology any more, so we now need to extract GitHub accounts as well.

I'm not sure if having multiple "handles" for the same maintainer is okay. If it isn't, replace `if` with `elif`.


AMDmi3:
Thanks, though I've decied to make it into `elif` not to duplicate maintainers unnecessarily.

PR #1310, opened 2023-02-16
aports parser: ignore _git versions

ptrcnull:
all versions with `_gitYYYYMMDD` should be ignored, as they're aports-only identifiers for versions using git commits instead of stable releases/tags

### DIFF
--- a/repology/parsers/parsers/aports.py
+++ b/repology/parsers/parsers/aports.py
@@ -52,6 +52,9 @@ class ApkIndexParser(Parser):
                 pkg.add_name(pkgdata['o'], NameType.APK_SMALL_O)
                 pkg.set_version(pkgdata['V'], _normalize_version)
 
+                if re.match('.*_git[0-9]{8}', pkg.version):
+                    pkg.set_flags(PackageFlags.IGNORE)
+
                 pkg.set_summary(pkgdata['T'])
                 pkg.add_homepages(pkgdata['U'])  # XXX: split?
                 pkg.add_licenses(pkgdata['L'])


ptrcnull:
btw, hinted by `# XXX: move to rules` in the nix parser, i looked at the repology-rules repo

and it looks like this rule is [already in place][1], but clearly doesn't work:

image


[1]: https://github.com/repology/repology-rules/blob/43e9bdfddb267d9c69c022a2047789dd0cba692a/899.version-fixes.global.yaml#L69

AMDmi3:
It works as expected. Outdated versions are not subject to ignore rules.

nekopsykose:
that specific software, cln, for which 1.3.6 is marked as newest, released 1.3.6 in december 2019 https://www.ginac.de/CLN/cln.git/  

20211105 is.. newer, not outdated.

nekopsykose:
i'd assume then it's parsed as 0 + git, so 0 is older than 1.3.6, but isn't it better to just always ignore on _git?

ptrcnull:
> It works as expected. Outdated versions are not subject to ignore rules.

it doesn't look like it:

- https://repology.org/project/git-lfs/versions, `2.12.0.r43.g68a8f8a4` ignored, latest is `3.3.0`
- https://repology.org/project/texstudio/versions, `3.0.4.r8.g829deb6b` ignored, latest is `4.5.1`
- https://repology.org/project/gcc/versions, `10.5.9999` ignored, latest is `12.2.0`

and those are only a few examples, so clearly outdated versions are still subject to ignore rules

AMDmi3:
> i'd assume then it's parsed as 0 + git, so 0 is older than 1.3.6, but isn't it better to just always ignore on _git?

It is not. Normally the snapshot version would be based upon the latest release - that way if the release is latest it won't be marked as outdated (ignored status would work), and othewise it would be correctly outdated by the newer release.

AMDmi3:
> it doesn't look like it:

These are rolling statuses for packages which fetch from master branch instead of a fixed commit.

ptrcnull:
> These are rolling statuses for packages which fetch from master branch instead of a fixed commit.

that's fair, but there are also versions in Repology that are being ignored when they don't match the upstream versioning scheme; for example OpenWrt's `YYYY-MM-DD-commithash`, which often is actually outdated

it would be nice if Repology could at least ignore 0_git versions (which are a common practice in aports to reference a commit, regardless if the project has actual releases), as there's no way of knowing at first glance whether the version is outdated or not; the `0` here should not be treated as a version

AMDmi3:
> that's fair, but there are also versions in Repology that are being ignored when they don't match the upstream versioning scheme; for example OpenWrt's `YYYY-MM-DD-commithash`, which often is actually outdated, or macports `YYYY.MM.DD`

The only difference for these is that they compare greater than the real versions and thus cause incorrect latest version reporting.  But in fact the policy is the same in both cases: we don't provide any meaningful handling for made up versions, we pessimize these either way, we do not introduce rules to fix these.

So I suggest to use real versions in aports instead.

PR #1311, opened 2023-02-21
debian-janitor: Add {bullseye,bookworm} backports repositories

jelmer:


AMDmi3:
- both repos need shorter names to fit on badges
- bookworm backports is empty
- what's semantics of these? if they just copy packages from next debian release, we probably don't need these

AMDmi3:
Timeout, also repos are unfetchable.

PR #1317, opened 2023-03-18
Map Nix attribute path to srcname

rhendric:
As hypothesized in repology/repology-webapp#218.


AMDmi3:
Let's add `binname` to `NIX_PNAME` mapping as discussed in the very same issue, right away.

AMDmi3:
Okay, makes sense.

PR #1320, opened 2023-03-21
Add Amazon Linux 2023

stewartsmith:
Unfortunately I haven't tested that this works, as haven't quite set up a repology dev environment yet. It does *appear* that this should work.... I'm not sure if the sqlite repodata classes mentioned also deal with the yum repo XML data? If so, then everything should be fine - if not, then I can't actually remember if we're also/still creating the sqlite repo data or not....

### DIFF
--- a/repos.d/rpm/amazonlinux.yaml
+++ b/repos.d/rpm/amazonlinux.yaml
@@ -41,3 +41,23 @@
     - desc: Amazon Linux 2
       url: https://aws.amazon.com/amazon-linux-2/
   groups: [ all, production, rpm, amazon ]
+
+- name: amazon_2023
+  type: repository
+  desc: Amazon Linux 2023
+  statsgroup: Amazon
+  family: fedora
+  ruleset: [fedora, rpm, amazon]
+  minpackages: 2000
+  sources:
+    - name: core
+      fetcher:
+        class: RepodataSqliteFetcher
+        url: https://cdn.amazonlinux.com/al2023/core/mirrors/latest/SRPMS/mirror.list
+      parser:
+        class: RepodataSqliteParser
+        vertags: amzn
+  repolinks:
+    - desc: Amazon Linux 2023
+      url: https://aws.amazon.com/linux/amazon-linux-2023/
+  groups: [ all, production, rpm, amazon ]


Conan-Kudo:
You want `RepodataFetcher` and `RepodataParser` instead. Those use the XML repodata.

AMDmi3:
> You want `RepodataFetcher` and `RepodataParser` instead

No you don't.

However, repositories are required to provide links to package sources.

stewartsmith:
> However, repositories are required to provide links to package sources.

What is in addition to the SRPM repository here that's needed?

stewartsmith:
For AL1, AL2, and AL2023 we tend to point people towards the SRPMs, as we don't directly extract them and host them anywhere, or currently have all the package sources in git repos publicly accessible.

Is this a new requirement from when AL1 and AL2 were added?

AMDmi3:
You can say so, requirements get stricter over time. Older repos will be removed at some point because of this too. The idea is that a repository which says it packages a specific version but doesn't provide an easy way to see the recipe and check where it comes from and how it is packaged, is useless to repology and dangerous to users.

childnode:
stumbled over here too

same same but different at https://github.com/childnode/repology-updater/commit/34f8016b95ae357a584f0ec5544922f6e8a02ff3

not sure why this has been closed and what is the consequence of https://github.com/repology/repology-updater/pull/1320#issuecomment-1479350341 @AMDmi3 ?!
or 
if this is all we need and didn't saw any tests ... perhaps someone can guide me what we need to get this done ,)

myself directing to here as a user of renovate which uses repology as a datasource https://docs.renovatebot.com/modules/datasource/repology/ e.g. 

expected 

```Dockerfile
# renovate: datasource=repology depName=amazon_2023/git versioning=loose
ENV GIT_VERSION=2.39.2-1.amzn2023.0.1
```

while 
```
# renovate: datasource=repology depName=amazon_2/git versioning=loose
```

will not work because of the distribution specific version, does it?

or do I understand it correct and the way to go is to declare
```
ENV GIT_VERSION=2.39.2
```

(which is fine for `dnf install git-2.39.2` too)

but again, we will need a dedicated amazon_2023 entry for pointing to `https://cdn.amazonlinux.com/al2023/core/mirrors/latest/SRPMS/mirror.list` instead of `http://amazonlinux.us-west-2.amazonaws.com/2/core/2.0/SRPMS/mirror.list`, do we?

With kind regards,
~Marcel

AMDmi3:
I think I've explained it in https://github.com/repology/repology-updater/pull/1320#issuecomment-1478580844. Just a link to the spec file.

childnode:
Dear @AMDmi3 I appreciate your response but I'm not that deep in as you are and not sure I didn't understood your point of what exactly will get dangerous. As I understand, the current listed repositories are too specific and neither good to maintain nor good for comprehensible package versions where it is unclear which is the source for, thats why you prefer SRPM but not the distribution specific reflection. Thats ok to me, but from a users point of view, it get's weired to still use "old" al2 pointers for 2023 distribution update schematics.
So what would be the way you want to go if you decline al2023? Should we upcoming just have one amazonlinux identifier for?
The thing I'm not that in deep insight is, why is there a subversioning per distribution? Is there any "backporting" behaviour so let's say a al2023 openssh package 
What's the difference between openssh
8.7p1-8.amzn2023.0.4 and 8.7p1-8.amzn2023.0.6 ? https://docs.aws.amazon.com/linux/al2023/release-notes/all-packages-al2023-20230501.html
Does this have any security related backport background? If yes, just looking at the source 8.7p1 didn't changed but perhaps i should have a look for the update either because it has had security flaws?

With kind regards, Marcel

AMDmi3:
I haven't understood most of this. To be added to Repology, repositories should provide direct links to package recipes. It's that easy.

childnode:
ok, I think I now understand your point 
https://github.com/repology/repology-updater/pull/1320/commits/c1ac5ef0172c24658f9253280a14c2bd56ddbc9c or my derivate https://github.com/childnodeContribArch/repology-updater/commit/34f8016b95ae357a584f0ec5544922f6e8a02ff3

are missing s.th. like 
```
packagelinks:
    - type: PACKAGE_HOMEPAGE
      url: 'https://pkgs.alpinelinux.org/packages?name={binname|quote}{{pkgs_branch}}'
PACKAGE_SOURCES
```
where the best we can provide is https://docs.aws.amazon.com/linux/al2023/release-notes/all-packages.html

and 
```
packagelinks:
    - type: PACKAGE_SOURCES
      url: 'https://git.alpinelinux.org/aports/tree/{subrepo}/{srcname}{{git_branch}}'
    - type: PACKAGE_RECIPE
      url: 'https://git.alpinelinux.org/aports/tree/{subrepo}/{srcname}/APKBUILD{{git_branch}}'
```
where best we can provide is NOTHING ?!

I appreciate your strictness on the FOSS appeal but my point in the previous rant was: repology is a great source to improve security by automated security updates in DevOps (e.g. via renovate). It should not be limited to those sources you, the maintainer esteem - but that's my 50¢ and opinion on this as a private FOSS guy.

Unfortunately I'm not that in in that RPM / DNF package management ... perhaps there is a way to query out the right URLs needed for this and anyone else might create a valid pull request ;/

AMDmi3:
Repology is not supposed to solve your devops tasks, it's supposed to aggregate information from repositories, and that is not possible if repository does not provide information.

EwoutH:
Would Amazon Linux 2023 support be possible? The problem is that there is that Amazon doesn't provide an URL to package recipes right?

AMDmi3:
Right.

PR #1329, opened 2023-04-22
Add Leap 15.3 and 15.4 to Packman

lgbaldoni:


AMDmi3:
Links to package recipes (e.g. spec files) are required for new repositories. Please provide these for packman. Also the patch contains invalid date in `valid_till`

PR #1332, opened 2023-05-23
Add Pardus 23

fthaltun:
Pardus 23 version has a separate repository from Debian, unlike previous versions. In this version, Pardus packages are offered separately without merging with Debian repositories.

### DIFF
--- a/repos.d/deb/pardus.yaml
+++ b/repos.d/deb/pardus.yaml
@@ -147,3 +147,29 @@
     - desc: Pardus home
       url: https://www.pardus.org.tr/
   groups: [ all, production, pardus ]
+
+- name: pardus_23
+  type: repository
+  desc: Pardus 23
+  statsgroup: Debian+derivs
+  family: debuntu
+  ruleset: [debuntu,pardus]
+  color: "ffcc00"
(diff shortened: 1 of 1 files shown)


repos.d/deb/pardus.yaml
```diff
@@ -147,3 +147,29 @@
...
   groups: [ all, production, pardus ]
+
+- name: pardus_23
+  type: repository
+  desc: Pardus 23
+  statsgroup: Debian+derivs
+  family: debuntu
+  ruleset: [debuntu,pardus]
+  color: "ffcc00"
+  minpackages: 5000
```
AMDmi3, line 158: It only has 11, so lets please change this to 1 or more if you're sure it won't contains less packages. The build failure is unrelated is should go away if you rebase.

AMDmi3 (changes requested):

repos.d/deb/pardus.yaml
fthaltun: We cannot give an exact number at the moment, we can give an exact number in about 2 months. Now, if there is no problem, we can say 1. Thanks.

PR #1336, opened 2023-06-03
Updates for ArchLinux repo changes

daenney:
* Arch repositories have moved to Gitlab under the packages/ tree
* Remove www. prefix as Arch uses the bare domain to avoid the redir badges

I'm not entirely sure if this is correct/complete, but I think this is most of it. I'm mostly unsure about the linkformatter but there seems to be 3 subrepos now in the UI:
* core, for example: https://archlinux.org/packages/core/x86_64/acl/
* extra, for example: https://archlinux.org/packages/extra/x86_64/0ad/
* multilib, for example: https://archlinux.org/packages/multilib/x86_64/lib32-aalib/

I think that means that the subrepo remapping that linkformatter was doing is no longer necessary?


AMDmi3:
Everything is correct, thank you!

PR #1337, opened 2023-06-04
Add Managarm x86_64

ElectrodeYT:


repology/parsers/parsers/managarm.py
```diff
@@ -0,0 +1,61 @@
...
+
+        for pkgname, props in plist_index.items():
+            pkg = factory.begin(pkgname)
+            if not props['pkgver'].startswith(pkgname + '-'):
+                pkg.log('pkgver is expected to start with package name', severity=Logger.ERROR)
+                continue
+
+            # Managarm packages have one unique name, and while the recipe files are sorted into seperate
+            # directories, the package names themselves are the only unique identifer that a package has.
+            pkg.add_name(pkgname, NameType.MANAGARM_NAME)
```
AMDmi3, line 47: If only a single name is used to refer to the package, one of `GENERIC_SRC_NAME`, `GENERIC_BIN_NAME`, `GENERIC_SRCBIN_NAME` should suffice.

repology/parsers/parsers/managarm.py
```diff
@@ -0,0 +1,61 @@
...
+import plistlib
+from typing import Iterable
+
+from repology.logger import Logger
+from repology.packagemaker import NameType, PackageFactory, PackageMaker
+from repology.parsers import Parser
+from repology.parsers.maintainers import extract_maintainers
+from repology.parsers.versions import VersionStripper
+
+# Managarm uses XBPS, however as the naming convention is different, and managarm lacks some package attributes
```
AMDmi3, line 30: I'd expect this in the PR comment, not in the source

repos.d/managarm.yaml
```diff
@@ -0,0 +1,28 @@
+###########################################################################
+# Managarm
+###########################################################################
+
+- name: managarm_x86_64
```
AMDmi3, line 5: Drop x86_64 part, please. For void it's a legacy artifact, in fact all arches should reside in the single repository.

repos.d/managarm.yaml
```diff
@@ -0,0 +1,28 @@
...
+  type: repository
+  desc: Managarm x86_64
+  family: managarm
```
AMDmi3, line 8: You need to add the family to repology-schemacheck.py as well.
Add `ruleset: managarm` here please too, as the plan is to make these explicitly separate (`family` is used to group related repositories, and `ruleset` is used to create specific rules).

repos.d/managarm.yaml
```diff
@@ -0,0 +1,28 @@
...
+  color: "97bde3"
+  minpackages: 240
+  default_maintainer: fallback-mnt-managarm@repology
+  sources:
+    - name: [ x86_64 ]
```
AMDmi3, line 13: If it's the only arch, you don't need [] and `{source}` substitutions later. `subrepo` may be removed as well.

repos.d/managarm.yaml
```diff
@@ -0,0 +1,28 @@
...
+      subrepo: '{source}'
+  repolinks:
+    - desc: Managarm home
+      url: https://managarm.org
+    - desc: Managarm build server
+      url: https://builds.managarm.org
+  packagelinks:
+    - type: PACKAGE_RECIPE
+      url: 'https://github.com/managarm/bootstrap-managarm'
+  groups: [ all, production, managarm ]
```
AMDmi3, line 28: You won't need `managarm` here as repository would be named so.

repos.d/managarm.yaml
```diff
@@ -0,0 +1,28 @@
...
+        class: ManagarmPlistParser
+      subrepo: '{source}'
+  repolinks:
+    - desc: Managarm home
+      url: https://managarm.org
+    - desc: Managarm build server
+      url: https://builds.managarm.org
+  packagelinks:
+    - type: PACKAGE_RECIPE
+      url: 'https://github.com/managarm/bootstrap-managarm'
```
AMDmi3, line 27: This is supposed to refer to a specific package recipe, usually an url template with `{srcname}` substitution.

repology/parsers/parsers/managarm.py
AMDmi3: Also note that flake8 requires 2 blank lines between imports and class def.

repology/parsers/parsers/managarm.py
```diff
@@ -0,0 +1,61 @@
...
+# compared to Void Linux, we need to use a different parser.
+class ManagarmPlistParser(Parser):
+    def iter_parse(self, path: str, factory: PackageFactory) -> Iterable[PackageMaker]:
+        normalize_version = VersionStripper().strip_right_greedy('_')
+
+        with open(os.path.join(path, 'index.plist'), 'rb') as plistfile:
+            plist_index: dict[str, dict[str, str]] = plistlib.load(plistfile, fmt=plistlib.FMT_XML)
+
+        for pkgname, props in plist_index.items():
+            pkg = factory.begin(pkgname)
```
AMDmi3, line 40: While here, please change it to `with factory.begin(pkgname) as pkg`. I should really clean legacy idioms out of existing parsers..

AMDmi3 (changes requested):

repos.d/managarm.yaml
ElectrodeYT: At the moment, a package recipe file can not be generated based on information available in the XBPS repo. I'll see if I can make something work, but at the moment recipe files are a group of .yaml files, named (approximately) by which gentoo package category they would belong to.

no92: I don't think that we would be able to do this with the setup we currently have. Best we could do with that is to link to the correct file, which contains a number of recipes (all of that gentoo-style category).

ElectrodeYT: Maybe we could also have links to the build log of the package?
It wouldn't be impossible to have XBPS add a link to the package metadata / redirect to the line in the github repo either, although that would all probably only begin working when packages are rebuilt (which, to be fair, happens quite often) and/or require updates to XBPS

AMDmi3: > I don't think that we would be able to do this with the setup we currently have

Unfortunately, recipe links are required.

> Maybe we could also have links to the build log of the package?

Sure, this is always welcome.

AMDmi3:
Timeout

PR #1343, opened 2023-07-25
Exherbo updates

tombriden:
* Addition of a PACKAGE_RECIPE url that's now in our packages.json
* Update urls to our new home at exherbolinux.org

### DIFF
--- a/repology/parsers/parsers/exherbo.py
+++ b/repology/parsers/parsers/exherbo.py
@@ -18,6 +18,7 @@
 import re
 from typing import Iterable
 
+from repology.package import LinkType
 from repology.package import PackageFlags
 from repology.packagemaker import NameType, PackageFactory, PackageMaker
 from repology.parsers import Parser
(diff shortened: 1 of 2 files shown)


AMDmi3:
Right about time -  I was going to remove broken repository.
- Recipe links are broken for gitlab, these miss `/-/` before `/blob/`
- Would be nice to have links to raw recipes as well
- `LinkType` missing in `from repology.package import`

tombriden:
> Right about time - I was going to remove broken repository.
> 
> * Recipe links are broken for gitlab, these miss `/-/` before `/blob/`
> * Would be nice to have links to raw recipes as well
The gitlab url does work without the `/-/`, which meant I could build the link in the same way for the gitlab and github repos; but yeah, it's probably best to use the correct url so have updated that and added the raw recipe urls too

> * `LinkType` missing in `from repology.package import`
oops, that's what i get for not testing it first :)

AMDmi3:
> The gitlab url does work without the

Right, that was mispaste on my side. Anyway, is it possible to add raw recipe links as well?

tombriden:
> > The gitlab url does work without the
> 
> Right, that was mispaste on my side. Anyway, is it possible to add raw recipe links as well?

Yep, they're in there now

tombriden:
is this good to merge now or...?

tombriden:
Hi @AMDmi3, is there anything more you need in this MR? I'm keen to get us back up and syncing with repology

AMDmi3:
No, it's perfect, thank you!

PR #1348, opened 2023-08-31
Process package recipe and license information for Cygwin

jon-turney:
Add links to package recipies, to make a start at getting rid of the scary warning at https://repology.org/repository/cygwin

### DIFF
--- a/repology/parsers/parsers/cygwin.py
+++ b/repology/parsers/parsers/cygwin.py
@@ -18,7 +18,7 @@
 import re
 from typing import Iterable
 
-from repology.package import PackageFlags
+from repology.package import LinkType, PackageFlags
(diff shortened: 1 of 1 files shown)


jon-turney:
Note: these links should exist for most, but not all packages (especially old ones).  If I've understood correctly, repology does check these exist, so I don't need to.

Otherwise I guess I can put the link (or flag that it exists) into the Cygwin's JSON package dump and conditionally add the link in the parser...

jon-turney:
Assuming the recipe URL is totally predictable turns out not to work in certain (historical) edge cases, so I've just added it to the JSON dump, and the changes here extract it from that.

Also add extracting license information, which we've had available for a while.

PR #1351, opened 2023-09-22
Update Solus recipe repo information to new GitHub monorepo

Staudey:
Solus has changed their package recipe repo from a self-hosted phabricator instance with split package repositories to a GitHub monorepo.

WIP Note: There is a special case for packages starting with "py" which is not considered in this initial PR draft. These have their own folder in the repo. I would need something like a `'pyorfirstletter': lambda x: x.lower()[:2] if x.lower().startswith('py') else x.lower()[0]` filter, but I'm unsure whether I should add that directly to the filter section of `repology/linkformatter.py` or somewhere else entirely. Input welcomed!

### DIFF
--- a/repology/linkformatter.py
+++ b/repology/linkformatter.py
@@ -28,6 +28,7 @@ _FILTERS: dict[str, Callable[[str], str]] = {
     'lowercase': lambda x: x.lower(),
     'first_letter': lambda x: x.lower()[0],
     'lib_or_first_letter': lambda x: x.lower()[:4] if x.lower().startswith('lib') else x.lower()[0],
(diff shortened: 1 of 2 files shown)


AMDmi3:
>  I would need something like a 'pyorfirstletter'

Yes, that's exactly what should be done - the case is very similar to `libfirstletter`. I've just renamed that to `lib_or_first_letter` for readability, you can use `py_or_first_letter`.

Staudey:
@AMDmi3 Thank you! I've done just that now. Should be good to go, unless I've missed something  (it's quite late here :sweat_smile:)

PR #1355, opened 2023-10-01
Add Debian Backports sloppy

daissi:
While packages in backports are based on those from release + 1 (the next direct release), backports-sloppy is based on release + 2.

For instance,  the list of packages provided with bullseye-backports-sloppy is available at https://backports.debian.org/bullseye-backports-sloppy/overview/


AMDmi3:
Merged through manual rebase.

PR #1356, opened 2023-10-02
Apertis: add more repolinks and packagelinks

daissi:
Apertis repositories pages complain about missing links, especially about links to package sources. This PR should fix that.

### DIFF
--- a/repos.d/deb/apertis.yaml
+++ b/repos.d/deb/apertis.yaml
@@ -29,6 +29,15 @@
   repolinks:
     - desc: Apertis home
       url: https://www.apertis.org/
+    - desc: Apertis GitLab
+      url: https://gitlab.apertis.org/
+    - desc: Apertis Quality Assurance reports
+      url: https://qa.apertis.org/
+  packagelinks:
+    - type: PACKAGE_SOURCES
+      url: 'https://gitlab.apertis.org/pkg/{srcname}/-/tree/apertis/{{v}}'
+    - type: PACKAGE_BUILD_STATUS
+      url: 'https://gitlab.apertis.org/pkg/{srcname}/-/pipelines?ref=apertis/{{v}}'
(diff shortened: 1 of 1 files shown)


repos.d/deb/apertis.yaml
```diff
@@ -29,6 +29,15 @@
   repolinks:
     - desc: Apertis home
       url: https://www.apertis.org/
+    - desc: Apertis GitLab
+      url: https://gitlab.apertis.org/
+    - desc: Apertis Quality Assurance reports
+      url: https://qa.apertis.org/
+  packagelinks:
+    - type: PACKAGE_SOURCES
+      url: 'https://gitlab.apertis.org/pkg/{srcname}'
```
AMDmi3, line 38: This probably needs to be tied to a branch, e.g. something like `https://gitlab.apertis.org/pkg/{srcname}/-/tree/apertis/{{v}}`

repos.d/deb/apertis.yaml
```diff
@@ -29,6 +29,15 @@
...
+    - type: PACKAGE_BUILD_STATUS
+      url: 'https://gitlab.apertis.org/pkg/{srcname}/-/pipelines?ref=apertis/v{{version}}'
```
AMDmi3, line 40: This seem to also refer to a branch name with suffix, so probably also `ref=apertis/{{v}}'`

AMDmi3 (changes requested):

repos.d/deb/apertis.yaml
daissi: Thanks, fixed.

repos.d/deb/apertis.yaml
daissi: Fixed

AMDmi3 (approved):

PR #1362, opened 2023-11-18
RepodataFetcher: apply zstd decompression to URL ending in .zst

andreasstieger:
as seen for openSUSE


AMDmi3:
Should'n this include repo configs update?

andreasstieger:
I noticed some of the repos do not specify a compression

AMDmi3:
Oh, I get it now. Perfect, thank you!

PR #1363, opened 2023-11-19
Fix yacp packages.json

fd00:
yacp packages.json hosting site (OSDN) became unstable, so the distribution source has been changed.


AMDmi3:
Url does not exist, the repsitory is now broken.

fd00:
Sorry, I created a sourceforge project and account, but it seems to have been disabled. (Perhaps it was determined that I have no track record of use).

I will try to come up with an alternative.

PR #1373, opened 2024-01-18
ConanCenter: Parse conanfile.py attributes, ignore cci. versions

valgur:
Extracts additional useful attributes from `conanfile.py` files using Python's `ast` module. Only loads simple literal values, no code is executed or evaluated.

Assigns the extracted attributes:
* description -> summary,
* homepage -> upstream_homepage,
* license -> licenses,
* topics -> categories (currently disabled since only the first topic would get loaded to the DB),
* package_type as an extra field (the "library" and "application" distinction might be useful).

Also added a rule to mark versions starting with `cci.` as "noscheme" as [ConanCenter uses these for snapshot versions](https://github.com/conan-io/conan-center-index/blob/66ef93e932c2bee07af648557f08777b8d622dd1/docs/adding_packages/conanfile_attributes.md#conancenter-specific-releases-format). I hope that's the correct label.

I tested it locally with
```
./repology-update.py --initdb
./repology-update.py --fetch --fetch --parse --database --postupdate conancenter
```
and confirmed that everything is working as expected. All recipes were parsed without issues.

### DIFF
--- a/repology/parsers/parsers/conan.py
+++ b/repology/parsers/parsers/conan.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with repology.  If not, see <http://www.gnu.org/licenses/>.
 
+import ast
 import os
 from collections import defaultdict
 from dataclasses import dataclass
@@ -22,6 +23,7 @@ from typing import Any, Callable, Iterable, Iterator
 
 import yaml
 
+from repology.package import LinkType, PackageFlags
 from repology.packagemaker import NameType, PackageFactory, PackageMaker
 from repology.parsers import Parser
 from repology.parsers.walk import walk_tree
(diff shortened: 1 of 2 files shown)


AMDmi3:
Sorry, there's no way we will parse code is if it were plain old data. This would break as soon as somebody adds f-string or variable reference for convenience.

AMDmi3:
> Any non-literal values (f-strings, string concatenation with +) will simply be ignored

That'll leave us with incomplete data which is not acceptable.

> ConanCenter does not allow dynamic values in these fields anyway.

Is there really strong documented policy on that? I doubt and even if there is I doubt it'll remain with repository growth. These recipes are code and code is meant to save human work, and that's done by code constructions. For instance, I can already see dict comprehension for `options` in `libxml2/all/conanfile.py`

> I suppose the ideal case would be a new endpoint or metadata dump from the repo?

Yes. It doesn't have to be live endpoint, regular dump is also fine. It's not strongly necessary though, conan provides minimal set of mandatory fields as it is. We can live perfectly fine without summaries and licenses, and even homepage urls as long as conandata download urls point to upstream.

valgur:
Thank you for elaborating. Much appreciated.

> That'll leave us with incomplete data which is not acceptable.

I checked the existing 1630 packages and there was not a single case currently where any of the relevant fields ('description', 'homepage', 'license', 'topics', 'package_type') was unparseable.
Also, is 0.1% incomplete data really more problematic than 100% missing metadata? That's probably rarer than the occurrence of simple errors in the metadata in the first place (such as inaccurate license tags).

> Is there really strong documented policy on that?

There isn't. But except for `license`, there isn't really a use case where the metadata might need to be dynamic, such as depending on the version number.

> I doubt and even if there is I doubt it'll remain with repository growth. These recipes are code and code is meant to save human work, and that's done by code constructions.

Conan just went through a long and somewhat painful 2.0 migration. Any disruptive changes are very unlikely for a couple of years, probably, and the basic metadata fields don't really look like a probable candidate for a change anyway.

> For instance, I can already see dict comprehension for options in libxml2/all/conanfile.py

And that's fine. Options and their defaults are out of scope for Repology anyway.

To recap. Am I understanding you correctly that the ConanCenter repo is basically too marginal / niche to be worth even the slightest risk of breakage or maintenance overhead, especially given that this metadata is non-essential?

AMDmi3:
> I checked the existing 1630 packages and there was not a single case currently where any of the relevant fields ('description', 'homepage', 'license', 'topics', 'package_type') was unparseable.

I've checked too with the same results.

> Also, is 0.1% incomplete data really more problematic than 100% missing metadata? That's probably rarer than the occurrence of simple errors in the metadata in the first place (such as inaccurate license tags).

Yes. Repology strives to be reliable and predictive service, not a stochastic scraper. We don't want to provide inconsistent data to maintainers and upstreams.

> basic metadata fields don't really look like a probable candidate for a change anyway

From my experience, if it's physically possible to extend POD with substitutions and code for convenience, it will happen. Even in most basic metadata. For instance, [examples](https://gist.github.com/AMDmi3/c21a10cd29e9e0cb9ea93a73366861ad) of substitutions in comment(summary) fields from FreeBSD ports. As soon as you want to distribute multiple flavors of the same thing, such as postgresql extension for multiple pg branches, or toolchain for multiple targets, or python modules for different python versions, or a bunch of similar things such as translation dictionaries or localization packages, you have to duplicate a lot of packages, and not to maintain that by hand you naturally reside to subclassing recipes or at least substituting distinctive parts. 

> To recap. Am I understanding you correctly that the ConanCenter repo is basically too marginal / niche to be worth even the slightest risk of breakage or maintenance overhead, especially given that this metadata is non-essential?

There's nothing marginal or niche about ConanCenter, same rules apply to all new code. [A lot](https://repology.org/docs/not_supported) of bigger or smaller repositories were rejected or delisted for the same reason. To recap,
- No repository regardless of size or popularity is worth a complex parser
  - Maintenance overhead.
  - Not reliable and in some cases not consistent.
  - Here, I also don't want to introduce dependency on Python.
  - We push repositories to publish machine readable data which could be used by other consumers.
- Additional metadata is really non-essential in this case.

PR #1374, opened 2024-01-18
Add an optional docker-compose.yml for easier dev setup

valgur:


AMDmi3:
Sory again, I'd like not to pollute repository with unportable stuff I don't use and don't maintain.

valgur:
@AMDmi3 Ok. You really hate having new contributors to the repo, don't you?

The Dockerfile also includes a few undocumented additional steps required to get the `libversion` extensions running, which I thought was useful.

Maybe you can tuck them away in some subfolder to lessen the "polluting" aspect.

And what exactly is "unportable" about Docker?

AMDmi3:
> You really hate having new contributors to the repo, don't you?

Not really. You shouldn't treat rejected contributions as not being welcome too. I've explained why I don't accept these.

> additional steps required to get the libversion extensions running

You mean `libversion.so.1` link? That has caught my eye, but I really don't think it's needed. Libraries are dynamically loaded by plain names, while versioning is usually only used in shared linking in conjunction with SONAME. Here, the library installed by `pgxs` the way it considers suitable, there's nothing in my Makefile which could affect it. I have it installed as `.so` both on FreeBSD and on Ubuntu and it works
```
% ls /usr/lib/postgresql/14/lib/libversion*; psql -U repology -Atc "select version_compare2('1.1', '1.2')"
/usr/lib/postgresql/14/lib/libversion.so
-1
```
Other extensions are installed the same way on both systems. Because it's not used in shard linking, you don't need ldconfig for `/usr/lib/postgresql/...` either.

> Maybe you can tuck them away in some subfolder to lessen the "polluting" aspect

Regardless of subfolder, if I don't maintain it, I don't want it in my repository.
I don't have a problem with keeping it in dedicated repository under repology org with write access for you if you wish.

> And what exactly is "unportable" about Docker?

It's not available on BSDs, for instance.

PR #1375, opened 2024-01-23
Add non-free-firmware for Kali

elboulangero:
References:
- https://www.kali.org/blog/non-free-firmware-transition/
- http://http.kali.org/dists/kali-rolling/non-free-firmware/

### DIFF
--- a/repos.d/deb/kali.yaml
+++ b/repos.d/deb/kali.yaml
@@ -13,7 +13,7 @@
   color: '32638f'
   minpackages: 27000
   sources:
-    - name: [ contrib, main, non-free ]
+    - name: [ contrib, main, non-free, non-free-firmware ]
       fetcher:
         class: FileFetcher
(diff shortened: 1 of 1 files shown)


AMDmi3:
It's empty for `kali-bleeding-edge`, is this expected? https://repology.org/log/3294157

elboulangero:
@AMDmi3 Yes it's empty, and it's expected that it _can_ be empty.

`kali-bleeding-edge` only contains packages that are automatically updated by the Kali bot, and I think there's no package in scope for the bot in kali-bleeding-edge, at the moment. This might change in the future.

I noticed that `non-free` is also empty for `kali-bleeding-edge`: http://http.kali.org/dists/kali-bleeding-edge/non-free/binary-amd64/

If that's a problem that a component is empty, feel free to remove `non-free` and `non-free-firmware` for kali-bleeding-edge then.

And thanks again for maintaining repology

PR #1376, opened 2024-01-24
terra: Add Terra rawhide

lleyton:
This adds the new Terra repo for Fedora rawhide :)


AMDmi3:
String literal should be quoted, otherwise probably good.

lleyton:
Alrightie, just did that.

lleyton:
heya, don't mean to bother, but anything else I should change?

lleyton:
Just added the Terra 40 repo as well, which is newly branched off

PR #1382, opened 2024-02-23
Remove development tag from fedora 39

jaymzh:
It's not in development anymore. :)

Signed-off-by: Phil Dibowitz <phil@ipom.com>


jaymzh:
BTW I _think_ this is why Fedora 39 is no longer getting updates in repology.

PR #1383, opened 2024-02-24
updated crux.yaml: update_period to 2h, valid_till date for 3.6

TimB87:
Hi @AMDmi3 

CRUX 3.6 is not receiving any updates anymore for quite some time, so we better not promote it anymore.

I also updated the sync interval to match the creation interval of the json file.

Cheers!

### DIFF
--- a/repos.d/crux.yaml
+++ b/repos.d/crux.yaml
@@ -11,7 +11,7 @@
   family: crux
   ruleset: crux
   minpackages: {{minpackages}}
-  update_period: 6h
+  update_period: 2h
   sources:
     - name: CRUX-ports-JSON
       fetcher:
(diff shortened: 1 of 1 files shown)


repos.d/crux.yaml
```diff
@@ -11,7 +11,7 @@
   family: crux
   ruleset: crux
   minpackages: {{minpackages}}
-  update_period: 6h
+  update_period: 2h
```
AMDmi3, line 14: Just remove it, please. As far as I remember, the reason for 6h was fetch problems which resulted in timeouts and blocked other updates, that doesn't seem to be the case now, all updates (both 304 and 200) are under 1 sec. Default update interval is 10 minutes (that's minimum, in practice it's still couple of hours) and we don't want to make it artificially bigger.

AMDmi3 (changes requested):

repos.d/crux.yaml
TimB87: Ah, I didn't know about that. I removed the line entirely.

mhoush: Just for a bit of clarification the 6h was chosen because that's how often the file on the CRUX side was updated at the time.

PR #1357, opened 2023-10-18
Add deepin 23 repositories

wineee:
- rename old deepin to `deepin 20`
- use deepin instead of Deepin (when using deepin as a standalone branding name, the first letter needs to be in lower-case)
- add  repositories(main, community, commercial) for `deepin 23`

PR #1384, opened 2024-02-26
Set sortname for 'deepin 20'

wineee:
cherry-pick from https://github.com/repology/repology-updater/pull/1357

### DIFF
--- a/repos.d/deb/deepin.yaml
+++ b/repos.d/deb/deepin.yaml
@@ -2,11 +2,10 @@
 # Deepin Linux
 ###########################################################################
 
-# Though we reserve place for deepin_{testing,stable} here, unstable
-# is currently the only repo (not taking precise/trusty into account)
 - name: deepin_unstable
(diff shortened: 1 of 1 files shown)


repos.d/deb/deepin.yaml
```diff
@@ -2,11 +2,9 @@
 # Deepin Linux
 ###########################################################################
 
-# Though we reserve place for deepin_{testing,stable} here, unstable
-# is currently the only repo (not taking precise/trusty into account)
-- name: deepin_unstable
+- name: deepin_20
```
AMDmi3, line 5: Please don't rename the repository as it's the same as removing and adding a new repository, loosing all history. You can add `sortname: deepin_20` here so it's ordered correctly against newer versions (#1357).

AMDmi3 (changes requested):

AMDmi3 (approved):

PR #1387, opened 2024-03-12
Add microsoft ubuntu and mono ubuntu repos

georgenicoll:
This change adds microsoft ubuntu and mono ubuntu repos for versions of the .net sdks etc.

The microsoft ubuntu repo doesn't provide a populated Sources/Sources.gz, so we need to parse Packages.gz.  Parsing this with the DebianPackagesParser was failing as microsoft ubuntu includes the Source field, so the DebianPackagesWithSourceFieldParser has been introduced to allow Packages files that include the Source.  If there is a reason that this shouldn't be done, let me know.


AMDmi3:
We do not support private repositories, sorry.

georgenicoll:
ok.  
Is it that the https://download.mono-project.com/repo/ubuntu/ repository is the problem?
Are you ok with https://packages.microsoft.com/ubuntu/?

AMDmi3:
I'm ok with neither.

felipecrs:
@AMDmi3 what does "private" mean exactly? Repositories whose source code is not open?

AMDmi3:
Not maintained by community and only serving limited set of packages.

PR #1399, opened 2024-04-21
ubuntu: Drop 23.04

jbicha:
It reached End of Life in January

https://wiki.ubuntu.com/Releases


AMDmi3:
We do not remove repositories

jbicha:
@AMDmi3 Did that policy change since September? https://github.com/repology/repology-updater/commits/master/repos.d/deb/ubuntu.yaml

There are a lot of missing Ubuntu End of Life Ubuntu releases. Do you want them restored?

AMDmi3:
There's no policy. Repositories have to be removed if they are no longer available from the mirrors.

PR #1410, opened 2024-06-18
Update variable in glaucus.py

firasuke:
Update build-time dependencies variable in glaucus.py


AMDmi3:
This was never correct, as it parsed dependency list into field for textual summary, I'm removing this line. Thanks for pointing to it.

PR #1416, opened 2024-07-29
Added MSVSphere Distribution, RHEL clone like AlmaLinux and Rocky

Tigro:
Hello! Please add MSVSphere Distribution. It based on RHEL (like AlmaLinux and Rocky) but has many improvements, such as multimedia support, classic desktop look, additional software, GOST encryption, and others.

At this time released only 9 (9.4) version of MSVSphere. 8 (8.10) version comming soon in August.

Site: https://msvsphere-os.ru
git: https://git.inferitos.ru/rpms

### DIFF
--- a/repos.d/rpm/centos.yaml
+++ b/repos.d/rpm/centos.yaml
@@ -321,4 +321,4 @@
 # See https://access.redhat.com/support/policy/updates/errata for EOL date (because it's a RHEL clone)
 # Release 8 coming soon
 #{{ msvsphere(8, minpackages=2500, valid_till='2029-05-31', subrepos=['AppStream','BaseOS','HighAvailability','NFV','PowerTools','RT','ResilientStorage','Extras']) }}
-{{ msvsphere(9, minpackages=2000, valid_till='2032-05-31', subrepos=['AppStream','BaseOS','CRB','HighAvailability','NFV','RT','ResilientStorage','Extras', 'LibreOffice']) }}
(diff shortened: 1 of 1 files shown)


AMDmi3:
Some of package sources/recipes links are invalid, for instance openvpn and xournalpp.

Tigro:
Yes. We use different branches in some git repositories: i9c, i9, i9ce, i9cf. Is any way to search proper branch from defined list? Or maybe some other way to fix this problem?

AMDmi3:
There should be a valid link to relevant spec for each package.

AMDmi3:
No, packagelinks are required.

Tigro:
Check this, please.

AMDmi3:
We probably don't want to include partial support for the distribution, as it would introduce confusion of why's a package not listed in Repology while it's indeed in the distribution. Also, code comments were not resolved.

If you have different branch naming policies for subrepositories, you can unroll `{% for sub in subrepos %}` template loop and provide dedicated `packagelinks` with corresponding branch names for each subrepository, see for example:

https://github.com/repology/repology-updater/blob/53a36858aa6f34abdf7cc2b5e84cd7f1bb694d61/repos.d/rpm/terra.yaml#L11-L47

Resummarising the requirements, we need all repositories with valid links to specs for all packages (not only should these exist but should also match package version, e.g. if there's `1.5.18-5.el9.inferit.2` version with inferit changes, spec should match that).

Tigro:
We have different branch naming policies inside one subrepository, like AlmaLinux.

AMDmi3:
You'll need to find a way to convey this required information then.

PR #1419, opened 2024-08-08
Add Artix repolinks, packagelinks, testing repos

CorySanin:


AMDmi3:
Do we really need duplicate packages from lib32?

CorySanin:
> Do we really need duplicate packages from lib32?

I'm just following [Arch's config](https://github.com/repology/repology-updater/blob/master/repos.d/arch/arch.yaml#L13). `lib32` is the Artix equivalent of `multilib` in Arch.

> Also, isn't there a way to point stable packages to a commit these were built from?

Not currently, no. I'll propose creating tags if it helps.

PR #1434, opened 2024-09-23
Add EoL date for mx linux

dreua:


AMDmi3:
Why not add dates for 21 and 23 right away?

dreua:
Yeah why not, done! Haven't thought of them because they are in the future tbh.

dreua:
I thought that this would remove MX Linux 19 from the badges with `exclude_unsupported` but it is still there. Is my thinking wrong or is there a mistake in this PR or is it just not applied yet?

Example: https://repology.org/badge/vertical-allrepos/pdfarranger.svg?columns=4&exclude_unsupported=1

dreua:
Great, thanks for the info! There's no hurry here, I just assumed it would be automatically deployed because the other changes went live so fast :rocket:

PR #1435, opened 2024-09-27
switch side default branch.

4fury-c3440d8:


4fury-c3440d8:
repology parser gets inactual package versions from master branch.

PR #1444, opened 2024-10-05
Add 22.03 LTS SP1, SP3, SP4 and 24.09 for openEuler

fundawang:
and remove distribs which are not supported now.

Thanks


AMDmi3:
Are there EoL dates for these repos?

fundawang:
> Are there EoL dates for these repos?
https://www.openeuler.org/en/download/archive/

PR #1445, opened 2024-10-22
sagemath.yaml: Add passagemath repository

mkoeppe:
https://github.com/passagemath/passagemath is a fork of sagemath


AMDmi3:
Parsing `upstream_url`s require additional logic to deal with variable substitutions, and this does not meet with our repository data format requirements. I see now that sagemath has the same problem leading to incorrect link parsing.

mkoeppe:
What can I do to help?

AMDmi3:
The best would be to provide repository data in machine readable format.

PR #1455, opened 2024-12-19
Add ROS support

nim65s:
Hi,

This is a first take at adding distributions from ROS: https://ros.org/

Adding more info (at least authors / maintainers / licenses) would require fetching package.xml from individual packages, but maybe I can automate that in a repository managed by the ROS project, which would provide flat file per distro version for repology to ingest.

### DIFF
--- a/repology/parsers/parsers/ros.py
+++ b/repology/parsers/parsers/ros.py
@@ -35,6 +35,12 @@ def ros_extract_recipe_url(key, /, *, url, tags, version, **kwargs):
         return url.removesuffix('.git') + f'/blob/{release}/package.xml'
     if 'gitlab' in url:
         return url.removesuffix('.git') + f'/-/blob/{release}/package.xml'
+    if 'bitbucket' in url:
+        # It is not possible to construct a bitbucket url for a file
+        # on a specific tag
+        return
+    err = f'ROS package {key} is neither on github, gitlab, or bitbucket'
+    raise RuntimeError(err)
 
 
 class RosYamlParser(Parser):
@@ -53,10 +59,6 @@ class RosYamlParser(Parser):
 
                 if 'version' not in release:
                     pkg.log(f'dropping {key}: has no version.', severity=Logger.ERROR)
-                    continue
-
-                if 'bitbucket' in release['url']:
-                    pkg.log(f'dropping {key}: RIP bitbucket', severity=Logger.ERROR)
                     continue
 
                 pkg.add_name(key, NameType.ROS_NAME)
--- a/repology/parsers/parsers/ros.py
+++ b/repology/parsers/parsers/ros.py
@@ -21,12 +21,7 @@ from repology.logger import Logger
 from repology.package import LinkType
 from repology.packagemaker import NameType, PackageFactory, PackageMaker
 from repology.parsers import Parser
-from yaml import load
-
-try:
-    from yaml import CLoader as Loader
-except ImportError:
-    from yaml import Loader
+from yaml import safe_load
 
 
 def ros_extract_recipe_url(key, /, *, url, tags, version, **kwargs):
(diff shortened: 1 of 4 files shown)


repology/parsers/parsers/ros.py
```diff
@@ -0,0 +1,83 @@
...
+
+                release = packagedata['release']
+
+                if 'version' not in release:
+                    pkg.log(f'dropping {pkg}: has no version.', severity=Logger.ERROR)
+                    continue
+
+                if 'bitbucket' in release['url']:
+                    pkg.log(f'dropping {pkg}: RIP bitbucket', severity=Logger.ERROR)
+                    continue
```
AMDmi3, line 65: I'm hesitant on adding a repo for which repology only sees a fraction of content. This is not consistent, leads to wtfs and possibly package flaps.

repos.d/ros.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - desc: ROS packages
+      url: https://index.ros.org/
+    - desc: {{name}} GitHub repository
+      url: https://github.com/ros/rosdistro/tree/master/{{name}}
+    - desc: {{name}} home
+      url: https://{% if version > 1 %}docs.ros.org/en{% else %}wiki.ros.org{% endif %}/{{name}}
+  groups: [ all, production, ros, ros{{version}} ]
+{% endmacro %}
+
+{{ ros(1, 'groovy', minpackages=200, valid_till='2014-07-31') }}
```
AMDmi3, line 36: For the note, it may make sense not to add EoLed repos of that would simplify code (e.g. by dropping legacy conditions for them).

AMDmi3 (changes requested):
Note there's only 10% intersection with other repos. Apart from that I don't see any major problems with package versions or names, but dropped packages due to lacking releases are a concern.

2024-12-19-172109_sel

repology/parsers/parsers/ros.py
nim65s: Oh, wait, humble is ros2, and the link actually work: https://bitbucket.org/qbrobotics/qbdevice-ros2-release/src/master/

nim65s: I can't find a way to provide a link to a file in a specific tag for bitbucket repo without fetching it :(

nim65s:
> Note there's only 10% intersection with other repos.

Yes, those repos are mostly specific to robotics applications, and mostly not packaged elsewhere.

They are available on Nix through https://github.com/lopsided98/nix-ros-overlay, and conda through https://robostack.github.io/, so this might change in the future.

> dropped packages due to lacking releases are a concern.

I'm not sure to understand what do you mean here. There are also dropped packages in older versions of other distributions too, right ? Anyways, it would be totally fine for me to just include the latest versions of ROS, if this may help. We currently have only 4 active ones, as Iron support has ended: https://docs.ros.org/

repology/parsers/parsers/ros.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def ros_extract_recipe_url(key, /, *, url, tags, version, **kwargs):
+    release = tags['release'].format(
+        package=key,
+        version=version,
+        upstream_version=version,
+    )
+
+    if 'github.com' in url:
```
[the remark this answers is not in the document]
AMDmi3: I'd prefer that to be `if/elif/else raise` then, so if another pattern appears the parsing fails and we can fix it right away.

repology/parsers/parsers/ros.py
AMDmi3: Sorry for misleading, the comment actually referred to all `continue`s here. We can live without a bunch of bitbucket links, but dropping a fraction of packages is a no-go.

repology/parsers/parsers/ros.py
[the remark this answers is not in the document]
AMDmi3: We can also pkg.add_links here right away. That way we can add RAW counterparts too.

AMDmi3:
> I'm not sure to understand what do you mean here. There are also dropped packages in older versions of other distributions too, right?

If you mean that some other parsers may skip packages - AFAIR there may be some, but we strive to either fix or remove these, not add new incomplete parsers.

repos.d/ros.yaml
nim65s: I did not have to write anything specific for these old things, and it seem to work so I included them. But I don't really need those.

nim65s:
Packages without releases are only available from source. This is often the case during the process of indexing a new package before the release repository can be created, eg. https://github.com/ros/rosdistro/pull/43915. Complete packages can turn into this state, for example when some ubuntu dependency is no longer available in new ubuntu version, eg: https://github.com/ros/rosdistro/pull/40091, but only on rolling, as other ones are tied on ubuntu versions.

Packages without version are those which fail to build on platform transition (same as latter case from above), eg. https://github.com/ros/rosdistro/pull/32036. In the same way, complete packages can turn into this state, but only on rolling.

AMDmi3:
We cannot add rolling then, incomplete repository content and flapping packages are show stoppers. I suggest to generate a complete dataset based on recipes and not affected by outside factors.

Pandapip1:
> Note there's only 10% intersection with other repos.

FWIW, I am trying to package the entirety of ROS in nixpkgs. The main issue is that it has a lot of ROS-specific dependencies with complex version restrictions making packaging it a bit of a pain. But it's entirely doable and I'd love someone to try to package it for the AUR too.

PR #1474, opened 2025-02-26
Parse URL/CPE-ID from OpenWrt packages

aparcar:
Those information been there for some time, use them.

### DIFF
--- a/repology/parsers/parsers/debian.py
+++ b/repology/parsers/parsers/debian.py
@@ -131,6 +131,10 @@ class OpenWrtPackagesParser(DebianSourcesParser):
             pkg.add_name(pkgdata['SourceName'], NameType.OPENWRT_SOURCENAME)
         pkg.set_arch(pkgdata['Architecture'])
         pkg.set_extra_field('path', '/'.join(pkgpath[2:]))
+        pkg.add_homepages(pkgdata.get('URL'))
+        if cpeid := pkgdata.get('CPE-ID'):
+            cpe_components = cpeid.split(':')
+            pkg.add_cpe(cpe_components[2], cpe_components[3])
 
         if pkgpath[2:4] == ['lang', 'python']:
             # All python modules are in lang/python, but not all of lang/python are python modules


repology/parsers/parsers/debian.py
```diff
@@ -131,6 +131,10 @@ def _extra_handling(self, pkg: PackageMaker, pkgdata: dict[str, str]) -> None:
             pkg.add_name(pkgdata['SourceName'], NameType.OPENWRT_SOURCENAME)
         pkg.set_arch(pkgdata['Architecture'])
         pkg.set_extra_field('path', '/'.join(pkgpath[2:]))
+        pkg.add_homepages(pkgdata.get('URL'))
```
AMDmi3, line 134: You sure? I don't see any `URL` fields in either subrepo for 23.05. `CPE-ID`'s are there though.

aparcar: Right, it was only added in https://downloads.openwrt.org/releases/24.10.0/targets/apm821xx/nand/packages/Packages.manifest, sorry

repology/parsers/parsers/debian.py
```diff
@@ -131,6 +131,10 @@ def _extra_handling(self, pkg: PackageMaker, pkgdata: dict[str, str]) -> None:
...
+        if cpeid := pkgdata.get('CPE-ID'):
+            cpe_components = cpeid.split(':')
+            pkg.add_cpe(cpe_components[3], cpe_components[4])
```
AMDmi3, line 137: Shouldn't that be 2 and 3, instead of 3 and 4?

AMDmi3 (changes requested):

repology/parsers/parsers/debian.py
AMDmi3, line 137: Also there's a single case of broken CPE-ID, 
```
CPE-ID: cpe:/a:lrzsz_project
```
from lrzsz in openwrt 2022. Can this be fixed in the repository or do we need a workaround here?

AMDmi3:
Ping?

repology/parsers/parsers/debian.py
aparcar: So the line is `CPE-ID: cpe:/a:netfilter:iptables`. If the parsing works right the content of `CPE-ID` should be `cpe:/a:netfilter:iptables`, which actually changes the right values to be 1 and 2, right?

aparcar:
The CI seems to be sad, anything I need to do about that?

```
Run sudo sed -i -e 's|peer|trust|' /etc/postgresql/$POSTGRESQL/main/pg_hba.conf
sed: can't read /etc/postgresql/15/main/pg_hba.conf: No such file or directory
```

AMDmi3:
No, just ignore it.

aparcar:
I updated the commit.

repology/parsers/parsers/debian.py
AMDmi3: No, why? [0] is `cpe`, [1] is `/a`, [2] is `netfilter` and [3] is `iptables`. We need 2 and 3.

AMDmi3:
https://github.com/repology/repology-updater/pull/1474#discussion_r1981957031

repology/parsers/parsers/debian.py
aparcar: Please feel free to drop 22.03 since it's no longer build. Even if I'd backport that patch, there is no build infrastructure to create new packages.

AMDmi3:
Thank you!

aparcar:
Please let me know once this is deployed, I'll then link it on our website

aparcar:
Please let me know once this is deployed so I can link it on our website

PR #1476, opened 2025-03-09
Remove outdated NixOS releases

tobim:
NixOS 24.11 is the only one that is not EOL. The Nix community normally sunsets a release roughly a month after the next one goes into GA.


AMDmi3:
Duplicate of #1242 and others

PR #1477, opened 2025-03-10
Add rosa 13

Survolog:
Add ROSA 13 repositories.

### DIFF
--- a/repos.d/rpm/rosa.yaml
+++ b/repos.d/rpm/rosa.yaml
@@ -71,6 +71,7 @@
 # Note: rosa2021.1+ does not have urpmi-compatible metadata
 # Note: rosa2019.1 was renamed to rosa2021.1, keeping old "name"
 - name: rosa_2019_1
+  sortname: rosa_2019_1
   type: repository
   desc: Rosa 2021.1
   statsgroup: Rosa
@@ -101,6 +102,7 @@
   groups: [ all, production, rosa ]
 
 - name: rosa_13
+  sortname: rosa_2023_1
   type: repository
   desc: Rosa 13
   statsgroup: Rosa
--- a/repos.d/rpm/rosa.yaml
+++ b/repos.d/rpm/rosa.yaml
@@ -71,7 +71,6 @@
 # Note: rosa2021.1+ does not have urpmi-compatible metadata
 # Note: rosa2019.1 was renamed to rosa2021.1, keeping old "name"
 - name: rosa_2019_1
-  sortname: rosa_2019_1
   type: repository
   desc: Rosa 2021.1
   statsgroup: Rosa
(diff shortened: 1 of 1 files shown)


mikhailnov:
+1

repos.d/rpm/rosa.yaml
```diff
@@ -100,6 +100,35 @@
       url: 'https://abf.io/import/{srcname}/raw/rosa2021.1/{srcname}.spec'
   groups: [ all, production, rosa ]
 
+- name: rosa_13
```
AMDmi3, line 103: Is it really 13 or 2023.1, why's discrepancy with the branch name?
If it's really 13, it needs a `sortname` to be ordered after 2021.

repos.d/rpm/rosa.yaml
```diff
@@ -100,6 +100,35 @@
...
+  type: repository
+  desc: Rosa 13
+  statsgroup: Rosa
+  family: rosa
+  ruleset: [rosa, rpm]
+  color: '3757a1'
+  minpackages: 20000
+  sources:
+    # note */updates and */testing repositories are empty ATOW
```
AMDmi3, line 112: Copypasted uncertainty is not acceptable. We need to know for sure whether there subrepositories are not used or should be included. We can fix this bit for 2021 along the way.

AMDmi3 (changes requested):

repos.d/rpm/rosa.yaml
Survolog: Thanks for the reaction.
Yes, it's really 13 now but git projects branch name on abf.io is rosa2023.1
https://mirror.rosa.ru/rosa/rosa13/iso/ROSA.FRESH.13/
https://rosa.ru/rosa-fresh-12/

I'll look for an example of how to use `sortname` and boost PR if someone doesn't do it sooner.

repos.d/rpm/rosa.yaml
AMDmi3: Great. If it's confirmed, you can just drop the comments.

repos.d/rpm/rosa.yaml
AMDmi3: You can use `sortname: rosa_2023_1`. In fact it can be anything which is lexographically greater than `rosa_2019_1`, but using the same format, which in addition corresponds to the branch name, feels bulletproof.

repos.d/rpm/rosa.yaml
```diff
@@ -71,6 +71,7 @@
 # Note: rosa2021.1+ does not have urpmi-compatible metadata
 # Note: rosa2019.1 was renamed to rosa2021.1, keeping old "name"
 - name: rosa_2019_1
+  sortname: rosa_2019_1
```
AMDmi3, line 74: Sortname is only needed for 13. Otherwise good.

AMDmi3 (changes requested):

PR #1480, opened 2025-03-15
Add CentOS Stream 10

sgvictorino:
and add Stream 9 NFV source


AMDmi3:
I've revamped centos stream configs, please recheck

https://github.com/repology/repology-updater/blob/master/repos.d/rpm/centos.yaml#L105-L156

AMDmi3:
Well I think we have plenty of time to correct it and it's better to have approximate EoL date than to have no EoL date at all.
Thanks for submission, it should appear on the site in a few hours.

PR #1492, opened 2025-04-30
Fix packages with nix teams not showing up

RossComputerGuy:
https://github.com/NixOS/nixpkgs/pull/394797 added a new `meta.teams` attribute to packages in nix, https://github.com/NixOS/nixpkgs/pull/400458 switched everything to use it. Without this fix, packages which use `meta.teams` do not show ownership on repology.

### DIFF
--- a/repology/parsers/parsers/nix.py
+++ b/repology/parsers/parsers/nix.py
@@ -219,6 +219,13 @@ class NixJsonParser(Parser):
                     else:
                         pkg.add_maintainers(extract_nix_maintainers(meta['maintainers']))
 
+                if 'teams' in meta:
+                    if not isinstance(meta['teams'], list):
+                        pkg.log('teams "{}" is not a list'.format(meta['teams']), severity=Logger.ERROR)
+                    else:
+                        for item in meta['teams']:
+                            pkg.add_maintainers(extract_nix_maintainers(item['members']))
+
                 if 'license' in meta:
                     pkg.add_licenses(extract_nix_licenses(meta['license']))
 


AMDmi3:
Is there a limit on a number of members in a team?

repology/parsers/parsers/nix.py
```diff
@@ -219,6 +219,16 @@ def iter_parse(self, path: str, factory: PackageFactory) -> Iterable[PackageMake
                     else:
                         pkg.add_maintainers(extract_nix_maintainers(meta['maintainers']))
 
+                if 'teams' in meta:
+                    if not isinstance(meta['teams'], list):
```
AMDmi3, line 223: You don't need data format checks, it should just fail on unexpected data. So this may probably be simplified to
```
for team in meta.get('teams', []):
    pkg.add_maintainers(extract_nix_maintainers(team['members']))
```

AMDmi3 (changes requested):

SigmaSquadron:
> Is there a limit on a number of members in a team?

No. The members list for teams can be arbitrarily long. https://github.com/NixOS/nixpkgs/blob/master/maintainers/team-list.nix#L1-L29

AMDmi3:
> > Is there a limit on a number of members in a team?
> 
> No. The members list for teams can be arbitrarily long. https://github.com/NixOS/nixpkgs/blob/master/maintainers/team-list.nix#L1-L29

We need a safeguard then to avoid N² for large teams maintaining lots of packages. Can either limit a list of maintainers from the team with e.g. 10, with a log if it's exceeded, or add a single maintainer derived from the team name instead.

winterqt:
N.B. this change isn't needed anymore as [`meta.maintainers` is now inclusive of `meta.teams`](https://github.com/NixOS/nixpkgs/pull/402991).

AMDmi3:
Should we close this then?

winterqt:
Yes, this patch shouldn't be needed anymore.

PR #1493, opened 2025-05-04
Remove default maintainer from Chimera

breakgimme:
chimera no longer has the concept of maintainers as of https://github.com/chimera-linux/cports/commit/910fcf1054398f95eb6b215dd702867d99edf250 so there's not much sense for a maintainer field to exist anymore

### DIFF
--- a/repos.d/alpine/chimera.yaml
+++ b/repos.d/alpine/chimera.yaml
@@ -11,7 +11,6 @@
   family: chimera
   ruleset: chimera
   minpackages: 6000
-  default_maintainer: fallback-mnt-chimera@repology
   sources:
     {% for subrepo in ['main', 'user'] %}
     - name: {{subrepo}}


AMDmi3:
First of all, chimera currently does not update, because the number of packages had recently [dropped](https://repology.org/log/8473178). Is this expected?

AMDmi3:
Then please confirm that urls in the repository config are correct and all subrepositories are listed. I suspect that many packages were recently moved into another subrepository.

breakgimme:
> Then please confirm that urls in the repository config are correct and all subrepositories are listed. I suspect that many packages were recently moved into another subrepository.


the repositories are correct but it seems like something happened with the apkindex and it just cuts out in the middle of the list, posted about it on irc just now so hopefully it gets resolved soon

AMDmi3:
Merged in dcba3e7c0e2ea91dc048ce7dda31c5a5dbf46b85

AMDmi3:
> but it seems like something happened with the apkindex and it just cuts out in the middle of the list, posted about it on irc just now so hopefully it gets resolved soon

It looks like it's the case again, could you please convey it?
```
Unexpectedly small number of packages: 5527 when expected no less than 10000
```
```
% tail -10 main
V:25.2.3.1-r1
T:Free office suite (Gujarati language pack)
U:https://www.libreoffice.org
L:GPL-3.0-or-later
A:x86_64
o:libreoffice
t:1745269652

P:libreoffice-lang_gug
V:25.2.3.1-r1
```
```
% tail -10 user
V:3.5.0-r2
T:Fast and accurate code counter (documentation)
U:https://github.com/boyter/scc
L:MIT OR Unlicense
A:x86_64
o:scc
t:1744558346

P:schilytools-cdrtools
V:2024.03.21-r0
```
Note that I'll have to disable repository in Repology if it can't reliably provide valid package data. I should also mention that both apkindex endpoints [reply with 500](https://repology.org/log/8514193) occasionally

breakgimme:
asked about it on irc again, not much more i can do unfortunately as i'm just a user/contributor

q66:
the issue seems to relate to how we run the apkbrowser app; when run through systemd service, the endpoint for index generation appears to time out and cut off

i'm currently away for 2 weeks but i'll fix it properly later when i am back; for now we have a workaround in place, it might not be entirely reliable but shouldn't cause persistent issues

PR #1509, opened 2025-06-16
Add: terra 42 repository

Owen-sz:


lleyton:
Looking at the notice on that page, I notice the broken redirect is for "wingpanel-indicator-nightlight-debugsource" on Terra Rawhide. However, this is to be expected as it was removed from the sources and [repo server](https://repos.fyralabs.com/terrarawhide/). Am I missing something here?

AMDmi3:
Current list of broken rawhide links:
https://gist.github.com/AMDmi3/0e38cc0ce934a66dc0c04e0432d193da

AMDmi3:
What do you mean by "delete"? These need to be fixed, not deleted.

hadikanani72 (approved):

AMDmi3:
Timeout. Feel free to resubmit when redirects are fixed.

Issue #1497, opened 2025-05-14
Nix maintainers limit


AMDmi3:
> Is there a limit on a number of members in a team?

No. The members list for teams can be arbitrarily long. https://github.com/NixOS/nixpkgs/blob/master/maintainers/team-list.nix#L1-L29

_Originally posted by @SigmaSquadron in https://github.com/repology/repology-updater/issues/1492#issuecomment-2846877575_

SigmaSquadron:
Yep, the main `meta.maintainers` list can also be arbitrarily long.

AMDmi3:
@SigmaSquadron @winterqt ping. If there's no better solution, I'll have to just trim or drop the maintainer list.

winterqt:
Hi @AMDmi3,

> These changes are not good, as these allow arbitrary long maintainer lists for arbitrary count of packages, resulting in N² of not too useful data. I'm currently removing team maintainers from maintainers, but that's not possible for stable branches as these do not have teams in meta.

This isn't any different from what we've done before -- we already had packages where `meta.maintainers` was, e.g. `teams.rust.members`, the changes discussed above are just a shortcut to that same outcome.

> IMO viable solution would be to not include team maintainers into maintainers, and reference team by a short name so it could be used as a maintianer itself.

This seems like a reasonable solution. Would you be receptive to a PR implementing this in Repology (for now)? Happy to do so.

winterqt:
> I'm currently removing team maintainers from maintainers, but that's not possible for stable branches as these do not have teams in meta.

Could you revert this change in the meantime, though? As discussed above, this isn't any different from before.

AMDmi3:
That's not a repology problem.

PR #1512, opened 2025-06-19
nix: Don't exclude team members

marcin-serwin:
This partially reverts 3fb04fa66f02377c26dab6fd747168a38d60d361.

As noted in https://github.com/repology/repology-updater/issues/1497#issuecomment-2899308391 the addition of `meta.teams` to `meta.maintainers` was just a return to the status quo before the introduction of separate field. In the long run it makes sense to handle teams as single entities but while it isn't implemented let's leave them as is, since some people rely on this (https://github.com/NixOS/nixpkgs/pull/402991#issuecomment-2853221104).

FWIW the largest team in nixpkgs counts 10 members so most of the team owned packages will likely still be under the limit.

### DIFF
--- a/repology/parsers/parsers/nix.py
+++ b/repology/parsers/parsers/nix.py
@@ -221,11 +221,6 @@ class NixJsonParser(Parser):
                 if 'maintainers' in meta:
                     maintainers = set(extract_nix_maintainers(meta['maintainers']))
 
-                    team_maintainers: set[str] = set()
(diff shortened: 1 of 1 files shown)


AMDmi3:
This breaks nix parsing as it would immediately throw from the immediately following condition.

AMDmi3:
I've bumped the limit to 20 for the time being.

PR #1520, opened 2025-07-14
Add deepin 25 repositories

BLumia:
Note: deepin 25's repo url path indeed still contain the word `beige` (which is the codename for deepin 23), it is not a typo.

### DIFF
--- a/repos.d/deb/deepin.yaml
+++ b/repos.d/deb/deepin.yaml
@@ -53,3 +53,32 @@
     - type: PACKAGE_SOURCES
       url: 'https://github.com/deepin-community/{srcname}'
   groups: [ all, production, deepin ]
+
+- name: deepin_25
+  type: repository
+  desc: deepin 25
+  statsgroup: Debian+derivs
+  family: debuntu
+  ruleset: [ debuntu, deepin ]
+  color: '949393'
+  minpackages: 8000
+  update_period: 1w
+  pessimized: "does not provide access (HTTP 404) to package sources (for instance, https://repology.org/link/https://github.com/deepin-community/linux)"
+  sources:
+    - name: [ main, community, commercial ]
+      fetcher:
+        class: FileFetcher
+        url: 'https://cdn-community-packages.deepin.com/deepin/beige/dists/crimson/{source}/source/Sources.gz'
+        compression: gz
+      parser:
+        class: DebianSourcesParser
+      subrepo: '{source}'
+  repolinks:
+    - desc: deepin home
+      url: https://www.deepin.org/
+    - desc: deepin open build service instance
+      url: https://build.deepin.com/
+  packagelinks:
+    - type: PACKAGE_SOURCES
+      url: 'https://github.com/deepin-community/{srcname}'
+  groups: [ all, production, deepin ]


BLumia:
I'm not an expert of system packaging and not directly working on package repo stuff, so I've reached out to the team working on community repo. The result I got is, while some of the missing source package issues can be resolved (e.g. dde-dock should simply be removed from the repo), it seems it's hard to address this issue for many other packages (e.g. rust and nodejs-based packages) since it have so many dependencies and it's hard to create a standalone repo for each of these packages on GitHub.

Since I'm not directly working on this and not able to tell if these are really addressable, I'll leave this PR open for a few days to see if we can have further discussion on this topic, if not, I'll close this PR.

cc @Zeno-sole @UTsweetyfish @xzl01

Thanks!

BLumia:
Hi, I'd like to confirm if I understand it correctly. The goals are:

1. For each package, we should *at least* be able to fetch its source (repo that contains a `debian/` folder) by visiting the link provided under `packagelinks`' `PACKAGE_SOURCES` configuration.
2. If `PACKAGE_SOURCES` contains invalid url, at least the package should come with a valid `Vcs-Browser` / `Vcs-Git` tag that point to a repo with `debian/` folder.

Is that correct? If so, my follow up questions are:

1. Can we have multiple `PACKAGE_SOURCES`?
   Most of deepin packages lives under two GitHub organizations: [`linuxdeepin`](http://github.com/linuxdeepin/) for first-party components and [`deepin-community`](http://github.com/deepin-community/) for upstream packages. While `dde-dock` is something we need to remove, its source live under the `linuxdeepin` organization. Other DDE/deepin core components also lives under that organization. If we can have multiple `PACKAGE_SOURCES`, we probably should do something like:
   ```
       - type: PACKAGE_SOURCES
         url: 'https://github.com/linuxdeepin/{srcname}'
       - type: PACKAGE_SOURCES
         url: 'https://github.com/deepin-community/{srcname}'
   ```
2. Is `point to debian repositories` not okay?
   While deepin does not use pre-built binaries from debian repo, we do import source from Debian, similar to what Ubuntu does. Thus `Vcs-Browser` / `Vcs-Git` is pointed to debian's repo unchanged (and also maintainer credits is kept as-is). I'd like to know if that's okay, and if not, what do you think that we should do to improve?
3. Is there a way to verify it by myself to see a list of packages having invalid source?
   I saw the README and it seems require `postgresql` installation and configuration. Is there a way to validate the configuration without setting that up? Or do I need to do the setup manually to validate it? If not, could you provide a full list of packages having invalid link issues?

Thanks!

AMDmi3:
Thank you for your interest in solving this issue!

> Hi, I'd like to confirm if I understand it correctly. The goals are:
> 
> 1. For each package, we should _at least_ be able to fetch its source (repo that contains a `debian/` folder) by visiting the link provided under `packagelinks`' `PACKAGE_SOURCES` configuration.
> 
> 2. If `PACKAGE_SOURCES` contains invalid url, at least the package should come with a valid `Vcs-Browser` / `Vcs-Git` tag that point to a repo with `debian/` folder.

Not quite.

Well let me first mention that the root of this requirement is not a formality, but a practical consideration, which is that any Repology user must be able to check where package comes from and how the package is built. For instance, to be able to get answers for cases like "I see a new version from repo XXX, but there's no such upstream tag. Is the version there made up, referencing a git commit and not an official release, or did I miss a release in some other place such as tarball on an official website, or does this package use a fork?" or "I see repo YYY packaging version 1.2.3 of package foo, but for me it doesn't build. How did they build it?".

Ideally, this information should be one click away, such as a direct link to package recipe (PKGBUILD, ebuild, FreeBSD port Makefile, homebrew formula etc.). For debian and some other packaging ecosystems it's a bit more complicated, so other kinds of links would suffice, such as a link to repository containing enough information for the case (such as suggested repository with debian/ contents only), or just a link to a web page containing further links all relevant info (like https://packages.debian.org/sid/source/hx pages), or a link to a service which redirects to a right place.

There are multiple ways to provide these links, for instance these can be parsed directly from the repository data (such as `Vcs-Browser`/`Vcs-Git` tags), or these can be generated from templates in repository configs, such as these PACKAGE_SOURCES links. In special cases, these can be generated by custom logic in the parser. You can't access the internet to check link validity on parsing stage though, so links generated either way must be valid.

So
- The goal is just to point a user to package sources.
- There are no restrictions on target resource format. Repository with debian/ contents is just one possible solution.
- Pointing to debian packages if deepin uses these unmodified is perfectly OK, as long as it can be trusted (e.g. there's no way that a debian package is modified by deepin, but someone forgets to update a link in the metadata).
- Link must be valid.
- Link can come either from from Vcs-* tags, or from template in repology repository config, or generated by parser code.
  
> Is that correct? If so, my follow up questions are:
> 
> 1. Can we have multiple `PACKAGE_SOURCES`?
>  Most of deepin packages lives under two GitHub organizations: [`linuxdeepin`](https://github.com/linuxdeepin/) for first-party components and [`deepin-community`](https://github.com/deepin-community/) for upstream packages. While `dde-dock` is something we need to remove, its source live under the `linuxdeepin` organization. Other DDE/deepin core components also lives under that organization. If we can have multiple `PACKAGE_SOURCES`, we probably should do something like:
> ```
> - type: PACKAGE_SOURCES
>   url: 'https://github.com/linuxdeepin/{srcname}'
> - type: PACKAGE_SOURCES
>   url: 'https://github.com/deepin-community/{srcname}'
> ```

If your intention is to provide two links, only one of which is valid, that's not a way to go (but technically having multiple links of one kind is supported). But if you provide valid links in tags, we can use that. Or we can generate one of these links in the parser as long as we can tell which one we want for a given package.

> 2. Is `point to debian repositories` not okay?

As mentioned above, these are okay as long as you really use unmodified debian sources for these packages.

> 3. Is there a way to verify it by myself to see a list of packages having invalid source?

Unfortunately, there's currently no way to see a list, and in fact we don't know all the invalid links because link checking for github links is subject to sampling (due to huge amount and a limited rate). I can run a query for you on demand, which can give some hints. Currently:
```
repology=> select url, ipv4_status_code from links where refcount > 0 and url like 'https://github.com/deepin-community/%' and ipv4_status_code not in (200, -2);
                              url                               | ipv4_status_code 
----------------------------------------------------------------+------------------
 https://github.com/deepin-community/dde-dock                   |              404
 https://github.com/deepin-community/gnome-desktop3             |              404
 https://github.com/deepin-community/rust-ansi-colours          |              404
 https://github.com/deepin-community/rust-array-macro           |              404
 https://github.com/deepin-community/rust-awaitable             |              404
 https://github.com/deepin-community/rust-beef                  |              404
 https://github.com/deepin-community/rust-blake2s-simd          |              404
 https://github.com/deepin-community/rust-chronoutil            |              404
 https://github.com/deepin-community/rust-clone-file            |              404
 https://github.com/deepin-community/rust-der-oid-macro         |              404
 https://github.com/deepin-community/rust-escargot              |              404
 https://github.com/deepin-community/rust-faccess               |              404
 https://github.com/deepin-community/rust-gif                   |              404
 https://github.com/deepin-community/rust-git-absorb            |              404
 https://github.com/deepin-community/rust-gix-worktree-state    |              404
 https://github.com/deepin-community/rust-hyperfine             |              404
 https://github.com/deepin-community/rust-iri-string            |              404
 https://github.com/deepin-community/rust-kmon                  |              404
 https://github.com/deepin-community/rust-koibumi-base32        |              404
 https://github.com/deepin-community/rust-kv-log-macro          |              404
 https://github.com/deepin-community/rust-lcms2                 |              404
 https://github.com/deepin-community/rust-libgweather           |              404
 https://github.com/deepin-community/rust-log4rs                |              404
 https://github.com/deepin-community/rust-lru                   |              404
 https://github.com/deepin-community/rust-privdrop              |              404
 https://github.com/deepin-community/rust-proc-quote            |              404
 https://github.com/deepin-community/rust-roaring               |              404
 https://github.com/deepin-community/rust-ruma-push-gateway-api |              404
 https://github.com/deepin-community/rust-rustls-pki-types      |              404
 https://github.com/deepin-community/rust-scala-native-demangle |              404
 https://github.com/deepin-community/rust-scopeguard-0.3        |              404
 https://github.com/deepin-community/rust-sequoia-ipc           |              404
 https://github.com/deepin-community/rust-simba                 |              404
 https://github.com/deepin-community/rust-speedate              |              404
 https://github.com/deepin-community/rust-ssh-format            |              404
 https://github.com/deepin-community/rust-starship-battery      |              404
 https://github.com/deepin-community/rust-stderrlog             |              404
 https://github.com/deepin-community/rust-stun-codec            |              404
 https://github.com/deepin-community/rust-symphonia-codec-aac   |              404
 https://github.com/deepin-community/rust-sysinfo               |              404
 https://github.com/deepin-community/rust-temptree              |              404
 https://github.com/deepin-community/rust-tui                   |              404
 https://github.com/deepin-community/rust-unicode-bom           |              404
 https://github.com/deepin-community/rust-universal-hash        |              404
 https://github.com/deepin-community/rust-vm-superio            |              404
 https://github.com/deepin-community/rust-widestring            |              404
 https://github.com/deepin-community/rust-wio                   |              404
(47 rows)
```

BLumia:
> The goal is just to point a user to package sources

Thanks for the explanation!

>> Can we have multiple `PACKAGE_SOURCES`?
>
> If your intention is to provide two links, only one of which is valid, that's not a way to go (but technically having multiple links of one kind is supported). But if you provide valid links in tags, we can use that. Or we can generate one of these links in the parser as long as we can tell which one we want for a given package.

I've asked our sysdev team, and by learning some debian packaging concept, I guess our first-party components (mostly, packages starts with `dde-` and `deepin-`) should have a valid `Homepage: ` field that point to its git repo page instead of deepin (distro)'s homepage. So if the work is done then it will point to the correct repo and we won't need to provide multiple `PACKAGE_SOURCES` links.

>> Is `point to debian repositories` not okay?
>
> As mentioned above, these are okay as long as you really use unmodified debian sources for these packages.

`PACKAGE_SOURCES` and `Vcs-Git`, Which one has higher priority? Most of our packages are unpatched, but if the package is patched, there must be a repo under the `deepin-community` GitHub organization.

Currently all rust-related packages are built with debian's `debcargo-conf` which is not patched by us and the `.dsc` file will point to the correct place ([`rust-array-macro`](https://cdn-community-packages.deepin.com/deepin/beige/pool/main/r/rust-array-macro/) for example):

```
Vcs-Browser: https://salsa.debian.org/rust-team/debcargo-conf/tree/master/src/array-macro
Vcs-Git: https://salsa.debian.org/rust-team/debcargo-conf.git [src/array-macro]
```

So if I understand it correctly, if the priority order is: `PACKAGE_SOURCES` > `Vcs-Git`/`Homepage` then it will be all fine:

1. Regular upstream packages will always have a repo contains its full source and `debian/` folder live under `deepin-community` GitHub organization. If the packages are patched by us, the repo will contain all patches.
2. DDE/deepin's first party projects will have a correct `Homepage:` field inside its `debian/control` file, which point to a valid git repo URL. (We will start working on this later)
3. Rust-based packages already have a valid `Vcs-Git` field.

One issue that I can image is for the 2nd one, these "fixes" will be included "later" instead of "right now", which need us to perform the fix, submit for testing and system integration and then push the update to the package repo.

> as long as you really use unmodified debian sources for these packages

And an additional question, is that needed to update the package's `Vcs-Git` field if it's patched by us, even if `PACKAGE_SOURCES` points to a repo that contains all patches?

We also checked Ubuntu as a reference, [`mesa`](https://launchpadlibrarian.net/799094751/mesa_25.0.7-0ubuntu0.25.04.1.dsc) package for example:

```
Homepage: https://mesa3d.org/
Standards-Version: 4.1.4
Vcs-Browser: https://salsa.debian.org/xorg-team/lib/mesa
Vcs-Git: https://salsa.debian.org/xorg-team/lib/mesa.git
```

...which is patched but also pointed to debian, and we both use the package's version number to indicate if the package is patched.

AMDmi3:
Sorry, on vacation. Will review it in a week.

AMDmi3:
> So if I understand it correctly, if the priority order is

There is no "priority order". Given a package Source entry, a valid link to its source should be generated, that is all. I don't see an algorithm by which we could choose between using Vcs-Git link or generating https://github.com/deepin-community/{srcname} link.

PR #1524, opened 2025-07-24
Add: terra 42

Owen-sz:
Broken links have been fixed.


AMDmi3:
No, there are still broken links for all terra versions.

Owen-sz:
Sorry about that, we had thought we got them all. Do you have an updated list of still-broken links?

Owen-sz:
The reason for asking about the list of links is to help us debug the issue with our tracking system. We recently deployed a patch from the input from the other PR, which we thought fixed the issues. It could be that there were some dangling entries from before the patched version was deployed, or that there’s still an issue, but it’s hard to know without seeing what entries are affected.

AMDmi3:
https://gist.github.com/AMDmi3/a25687ad250632b846bd5d00fd05c6f3
vencord-desktop looks to be rather fresh

PR #1526, opened 2025-08-01
Enable rosa13

Survolog:
At the moment, rosa13 is a live platform for the distribution of Rosa Fresh.
It is where the latest releases are released Rosa Fresh.


AMDmi3:
> I thought that I changed this line

I've had the same idea

PR #1532, opened 2025-09-04
Adding hydrapwk repo

me-joe:
name: hydrapwk
codename: yellowdog
repo url: http://repo.rstrike.my.id/yellowdog
component: main non-free
arch: amd64, arm64

Thank you! (Hopely not bother you)

### DIFF
--- a/repos.d/deb/hydrapwk.yaml
+++ b/repos.d/deb/hydrapwk.yaml
@@ -15,7 +15,7 @@
     - name: yellowdog/{{component}}/{{arch}}
       fetcher:
         class: FileFetcher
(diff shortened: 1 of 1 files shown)


me-joe (commented):
Wrong target

AMDmi3:
No package source links

me-joe:
> No package source links

For only Packages.gz not working?

me-joe (commented):
Replace packages.gz to sources.gz

AMDmi3:
That's not what I've requested. Repology can work with Packages, though Sources is strongly preferred. It's now broken in multiple places, for instance config still mentions arch, Sources are missing for `non-free` and `main` only contain one package.

What I've requested is a link to sources for each package, such as https://sources.debian.org/src/ansible/12.0.0~b5+dfsg-1/

Finally, why do we need this 20 package repository in repology?

PR #1539, opened 2025-10-06
Update openEuler with new 25.09 and prune outdated entries

ottok:
See commits for details.


fundawang:
> FYI @fundawang

lgtm

AMDmi3:
- The current policy is to remove EoLed repos 2 years after EoL date, please do not remove entries before that
- Updates to openEuler will not be accepted until fatal issues are fixes, see [repository page](https://repology.org/repository/openeuler_25_03)

ottok:
Thanks for the clarification. I am not involved openEuler so I can't fix
anything, and I don't even know where they track distro level bugs like
this, but maybe @fundawang can help?

PR #1545, opened 2025-11-04
pypi: Fix confusion between falsy and UPSTREAM_HOMEPAGE

TobiX:
This dropped all URLs of type "UPSTREAM_HOMEPAGE" before.

This fixes homepage links for projects which got rid of the legacy URL fields and only use modern metadata (see https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#urls).

### DIFF
--- a/repology/parsers/parsers/pypi.py
+++ b/repology/parsers/parsers/pypi.py
@@ -90,7 +90,7 @@ class PyPiCacheJsonParser(Parser):
 
                 if info['project_urls']:
                     for key, url in info['project_urls'].items():
-                        if (link_type := _url_types.get(key.lower())) and url != 'UNKNOWN':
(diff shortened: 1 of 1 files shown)


repology/parsers/parsers/pypi.py
```diff
@@ -90,7 +90,8 @@ def iter_parse(self, path: str, factory: PackageFactory) -> Iterable[PackageMake
 
                 if info['project_urls']:
                     for key, url in info['project_urls'].items():
-                        if (link_type := _url_types.get(key.lower())) and url != 'UNKNOWN':
+                        link_type = _url_types.get(key.lower())
```
AMDmi3: You can still check in the same expression. Also please check with `is not None`.

Issue #1499, opened 2025-05-22
Start removing EoLed repositories


AMDmi3:
Latest events such as a migration to larger VPS instance due to diskspace shortage and re-enabling the link checker, which showed that a lot of disk space and link checker capacity is taken by the old EoLed repositories. It's obvious that we can't keep these indefinitely, as it also affects database update times which grows with each repository release, repository configs complexity, which includes conditions for older repositories, and ruleset complexity for the same reason, and amount of scroll for project pages. There have been multiple requests to hide EoLed repos from the site by default.

The benefit of keeping these is less clear. My main use cases for it was to
- Show the (still) users of EoLed repositories how much these actually are outdated and vulnerable.
- Keep links to historic build recipes in case these are needed again.

The latter doesn't look valid enough
- It's unlikely that recipes for abandoned ancient software which is removed from the new repos should not be even needed; if some software is resurrected, it will likely need a recipe written from scratch; old recipes would unlikely be compatible too
- It doesn't solve preservation task anyway as packages disappearing from live repositories are just removed

The former still seems valid, but we don't plan to remove EoLed repos right away, and newer, but outdated and vulnerable enough repository releases should be representative enough.

So let's start by removing repos EoLed older than 2 years ago. I'd start specifically with repositories which frequent (thus a lot of) releases such as alpine, fedora and nix, and make exceptions for repositories with extended support or very extended third party support, such as centos. I'd also keep unique packaging ecosystems for now, as these may contain useful quirks and patches.

AMDmi3:
After cleanup
```
          name          |     years_ago      
------------------------+--------------------
 ataraxia               | 3.6518299419301116
 centos_6               | 4.4792272022040842
 centos_8               | 3.3942956953547692
 epel_6                 | 4.4792272022040842
 just-install           | 3.4600491200123034
 kiss_games             | 3.6819669282314815
 lmde_4                 | 2.7285422706972349
 sabayon                | 3.7723778871355911
 sabayon_for_gentoo     | 3.7723778871355911
 ubuntu_14_04           | 6.0819669282314815
 ubuntu_14_04_backports | 6.0819669282314815
 ubuntu_16_04           | 4.0655285720670979
 ubuntu_16_04_backports | 4.0655285720670979
(13 rows)
```

AMDmi3:
I've just given my considerations

PR #1546, opened 2025-11-05
Update rpmfusion current maintained version

kwizart:


AMDmi3:
Our current policy is to remove repositories 2 years after their EoL date. Has 2 years passed for removed repos? Please provide EoL dates if possible.

kwizart:
I think to have fixed  all issue. Please confirm.

The only remaining issue is likely gstreamer1-libav on el9, but this package  is already retired in our repository and will be removed from the repository on the next push. (It will occur over the week-end).

> So, what about EoL dates?

centos 7 was EOL on 2024/06 https://endoflife.date/centos
But still advertising it is a problem as we are likely going to move the packages elsewhere

Same for Fedora https://endoflife.date/fedora.

Also no updates are expected to be made, that's what's EOL means. So having theses releases displayed along supported ones is problematic IMHO.

> And another question, do I understand correctly that rpm fusion fedora rawhide correspond to master branches in package repositories?

Yes

kwizart:
> xine-plugin is the only one remaining. Probably a retired package too.

Fixed, thanks.

> we were keeping these indefinitely, https://github.com/repology/repology-updater/issues/1499 for more info.

About centos EOL, the RHEL 6 ELS only really applies to a very limited list of packages:

https://access.redhat.com/support/policy/updates/errata
https://access.redhat.com/articles/4997301

So I don't think this applies for general distro, and it specially not apply to our rpmfusion content.

But anyway, I will leave the older ones so they are discarded as your discretion.

Fixup....

kwizart:
I've fixed xine-plugin and the new list. If anything remains please let me know.
Thanks

AMDmi3:
Thank you! Only 8 urls remain
```
https://github.com/rpmfusion/ivsc-firmware/blob/f42/ivsc-firmware.spec
https://github.com/rpmfusion/ivsc-firmware/blob/f43/ivsc-firmware.spec
https://github.com/rpmfusion/ivsc-firmware/blob/master/ivsc-firmware.spec
https://github.com/rpmfusion/ivsc-firmware/tree/f42
https://github.com/rpmfusion/ivsc-firmware/tree/f43
https://github.com/rpmfusion/kffmpegthumbnailer/blob/f43/kffmpegthumbnailer.spec
https://github.com/rpmfusion/kffmpegthumbnailer/tree/f43
https://github.com/rpmfusion/ripcord/blob/master/ripcord.spec
```

kwizart:
Fixed, the remaining master branches will be fixed on the next push (by the end of the week). theses components were not properly retired.

PR #1550, opened 2025-11-20
Add deepin 25 repositories

wineee:
Use the same PACKAGE_SOURCES as Deepin 23; the only difference is that Deepin 23 might provide an older version of the package.


AMDmi3:
Duplicate of #1520

PR #1557, opened 2025-12-09
Update Fedora 41 eol date again

dreua:
Eol was postponed, sorry.
https://discussion.fedoraproject.org/t/fedora-41-eol-whoops/175073/4


dreua:
Thanks! I'll try to remember and send PRs only after making sure that the eol really happened in the future ;)

PR #1560, opened 2025-12-28
Add AlmaLinux 10 repository configuration

jonathanspw:


AMDmi3:
There are no links to package souces/.specs set for almalinux repositories, no new versions are accepted until that's fixed.

PR #1567, opened 2026-01-18
Implement MARKED_BROKEN flag for OpenBSD packages

Mayank7Bisnewar:
Body: This PR introduces a new package flag to handle OpenBSD ports marked as BROKEN.

Changes:
- Added PackageFlags.MARKED_BROKEN (1 << 21).
- Updated OpenBSD parser (openbsd.py) to set this flag when the Broken column is populated in sql ports .
- Modified : update_vulnerabilities.sql to exclude packages with this flag from vulnerability statistics.

### DIFF
--- a/repology/package.py
+++ b/repology/package.py
@@ -87,6 +87,7 @@ class PackageFlags:
     NOLEGACY: ClassVar[int] = 1 << 18
     OUTDATED: ClassVar[int] = 1 << 19
     RECALLED: ClassVar[int] = 1 << 20
+    MARKED_BROKEN: ClassVar[int] = 1 << 21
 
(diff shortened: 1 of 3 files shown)


AMDmi3:
Thanks for the elaboration.

- The logic in this change takes any `broken` item into account, while these are in fact limited with arch in most cases.
- I don't feel like vulnerability masking makes sense. Vulnerability in a dead code is still a vulnerability, and is a useful signal to users and maintainers. The intersection of broken & vulnerable is too small to affect prioritization in any way.

However, it does look sensible to hide (unconditionally) broken packages completely by not emitting these from the parser, with a corresponding log entry.

AMDmi3:
For the record, these changes were reviewed in #1568

PR #1576, opened 2026-02-27
Add Leap 16.0 and 16.1

hramrach:


AMDmi3:
I don't think I've understood any of that, but regardless unless the mirror problem is fixed there's no way to advance here.

adrianschroeter:
JFYI, there won't be an :Update project anymore . Leap 16.x is shipping it's update now as part of the main repository.

PR #1584, opened 2026-04-15
Security: Potential SQL injection via unsanitized SQL construction

tomaioo:
## Summary

Security: Potential SQL injection via unsanitized SQL construction

## Problem

**Severity**: `Medium` | **File**: `repology/parsers/sqlite.py:L24`

`iter_sqlite` builds SQL using string formatting for `table_expr` and `columns` (`SELECT {} FROM {}`). If either argument is derived from untrusted input (directly or indirectly via configuration), attackers can inject arbitrary SQL into the query.

## Solution

Avoid dynamic SQL where possible. For identifiers (table/column names), enforce strict allowlists or regex validation (e.g., `^[A-Za-z_][A-Za-z0-9_]*$`) and map user-facing values to predefined safe SQL fragments.

## Changes

- `repology/parsers/sqlite.py` (modified)

### DIFF
--- a/repology/parsers/sqlite.py
+++ b/repology/parsers/sqlite.py
@@ -15,11 +15,22 @@
 # You should have received a copy of the GNU General Public License
 # along with repology.  If not, see <http://www.gnu.org/licenses/>.
 
+import re
 import sqlite3
 from typing import Any, Iterable, Sequence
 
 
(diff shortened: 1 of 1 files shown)


AMDmi3:
Useless neuroslop

PR #1585, opened 2026-04-19
Add XRepo repository fetch support

luadebug:
Similiar to vcpkg and conan there is https://github.com/xmake-io/xmake-repo that provides C++ packages.

### DIFF
--- a/repology-schemacheck.py
+++ b/repology-schemacheck.py
@@ -134,6 +134,7 @@ families = [
     'wakemeops',
     'wikidata',
     'winget',
+    'xrepo',
     'yacp',
     'yiffos',
 ]
--- a/repology/fetchers/fetchers/xrepo.py
+++ b/repology/fetchers/fetchers/xrepo.py
@@ -1,0 +1,115 @@
+# Copyright (C) 2024 Repology contributors
+#
+# This file is part of repology
+#
+# repology is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# repology is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with repology.  If not, see <http://www.gnu.org/licenses/>.
+
+import json
+import re
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from typing import Any
+
+from repology.atomic_fs import AtomicFile
+from repology.fetchers import PersistentData, ScratchFileFetcher
+from repology.fetchers.http import do_http
+from repology.logger import Logger
+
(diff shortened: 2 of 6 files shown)


AMDmi3:
This tries to parse lua code with regular expressions. This is unfortunately not acceptable.

AMDmi3:
The problem was not in regular expressions, but in parsing lua source code. We will not parse source code in any way.

luadebug:
> The problem was not in regular expressions, but in parsing lua source code. We will not parse source code in any way.

I see, so that means we can not refer to xmake.lua files, but maybe some third source like some website. For example parsing https://xmake.microblock.cc/ can do a trick. Otherwise I am not sure that would be possible.

luadebug:
@AMDmi3 ready to review.

AMDmi3:
This approach is viable, however a show stopper is that no upstream urls are extracted (which is required), and there doesn't seem to be a way to extract these from the data:
- `homepage` (from `set_homepage`) is missing
- `urls` are not usable as there's no way to (reliably, which rules out checking for `.git` suffix) tell repository url from download url, also string interpolation (`$(version)` expansion) is not allowed.

AMDmi3:
Like I've just said, classifying urls by extension and string interpolation are not acceptable. API pagination is not acceptable either.

gistrec:
> This approach is viable, however a show stopper is that no upstream urls are extracted (which is required), and there doesn't seem to be a way to extract these from the data

I opened an upstream issue in xmake-repo to ask whether they can provide a generated machine-readable package manifest: https://github.com/xmake-io/xmake-repo/issues/9966

As I understand it, parsing `xmake.lua` is not an option here, and the current catalog source is close but still misses the fields needed by Repology: `homepage`, clearly separated repository/download URLs, and concrete URLs without `$(version)` interpolation

So maybe the best path is to keep this PR blocked until xmake-repo provides a proper generated index. If they agree to publish such a manifest, this fetcher could be simplified to consume that file instead of relying on Lua parsing or fragile URL classification

I think xrepo data would be valuable for Repology, so hopefully this can be solved upstream

gistrec:
Quick heads-up for context (not asking for action here): xmake-io/xmake-repo#9967 is preparing an upstream, machine-readable manifest specifically for Repology. Once merged it will be published at:

`https://xmake-io.github.io/xmake-repo/repology_packages_index.json`

It addresses the blockers raised earlier in this thread:

- **No Lua / source parsing** — generated via xmake's own loader (`package.load_from_repository`).
- **`repository_url` and `download_url` are separate fields** — repository URLs are normalized from `add_urls` (forge layouts) with a homepage fallback; download URLs are the first `add_urls` entry per package.
- **No `$(version)` interpolation in the manifest** — placeholders are resolved against the latest version on the generator side.
- **No pagination** — single flat JSON, ~1952 entries today.

If/when that lands, an `xrepo` fetcher could be reduced to a single HTTP GET + JSON parse, without the regex/heuristic logic in this PR.

gistrec:
FYI: I just opened #1594 with an alternative implementation based on the static index xmake-repo publishes at [repology_packages_index.json](https://xmake-io.github.io/xmake-repo/repology_packages_index.json) (updates every 2 hours)

AMDmi3:
Superseded by #1594

PR #1587, opened 2026-04-24
repos.d/deb/devuan.yaml: add Devuan 5.0 and 6.0

### LINKED ISSUE #1586: Update devuan to support 5 and 6
`repos.d/deb/devuan.yaml` supports 4.0 and unstable, but not 5 or 6.

I _think_ we just need additional stanzas in that file.

hlein:
Closes: https://github.com/repology/repology-updater/issues/1586

### DIFF
--- a/repos.d/deb/devuan.yaml
+++ b/repos.d/deb/devuan.yaml
@@ -1,50 +1,73 @@
 ###########################################################################
 # Devuan
-#
-# See release info on https://www.devuan.org/os/releases
 ###########################################################################
-- name: devuan_4_0
+
+{% macro devuan(version, codename, minpackages, valid_till=None) %}
+{% set subrepos2 = ['main', 'contrib', 'non-free'] %}
+{% set host = 'deb.devuan.org' %}
+
+- name: devuan_{{version}}
   type: repository
-  desc: Devuan 4.0 # Chimaera
+  desc: Devuan {{version}}
   statsgroup: Debian+derivs
   family: debuntu
-  ruleset: debuntu
+  ruleset: [debuntu, devuan]
   color: '88839c'
-  valid_till: 2026-08-15
-  minpackages: 27000
+  minpackages: {{minpackages}}
+  {% if valid_till %}
+  valid_till: {{valid_till}}
+  {% endif %}
   sources:
-    {% for sub1 in ['chimaera', 'chimaera-updates'] %}
-    {% for sub2 in ['main', 'contrib', 'non-free'] %}
+    {% for sub1 in [codename, codename + '-updates'] %}
+    {% for sub2 in subrepos2 %}
     - name: {{sub1}}/{{sub2}}
       fetcher:
         class: FileFetcher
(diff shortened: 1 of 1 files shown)


AMDmi3:
Ordering is reversed, and in fact it should be converted to a macro to avoid copypasta. See other debian repos for examples of such macros,

hlein:
> Ordering is reversed, and in fact it should be converted to a macro to avoid copypasta. See other debian repos for examples of such macros,

OK, redone to use macros based on `debian.yaml`, please take a look.

repos.d/deb/devuan.yaml
```diff
@@ -1,50 +1,73 @@
 ###########################################################################
 # Devuan
-#
-# See release info on https://www.devuan.org/os/releases
 ###########################################################################
-- name: devuan_4_0
+
+{% macro devuan(version, codename, minpackages, valid_till=None) %}
+{% set subrepos2 = ['main', 'contrib', 'non-free'] %}
+{% set host = 'deb.devuan.org' %}
```
AMDmi3, line 7: No need to have a variable for host, as it does not change depending or subrepository

repos.d/deb/devuan.yaml
```diff
@@ -1,50 +1,73 @@
...
+
+- name: devuan_{{version}}
   type: repository
-  desc: Devuan 4.0 # Chimaera
+  desc: Devuan {{version}}
   statsgroup: Debian+derivs
   family: debuntu
-  ruleset: debuntu
+  ruleset: [debuntu, devuan]
```
AMDmi3, line 14: There's no such ruleset, and we most likely don't need one

repos.d/deb/devuan.yaml
```diff
@@ -1,50 +1,73 @@
...
-        url: 'http://deb.devuan.org/merged/dists/{{sub1}}/{{sub2}}/source/Sources.gz'
-        compression: gz
+        url: 'http://{{host}}/merged/dists/{{sub1}}/{{sub2}}/source/Sources.xz'
+        compression: xz
       parser:
         class: DebianSourcesParser
       subrepo: {{sub1}}/{{sub2}}
     {% endfor %}
     {% endfor %}
+    {% set sub1 = codename + '-security' %}
```
AMDmi3, line 33: There's no need to put this into separate section, the body is the same. Just extend `sub1` loop on line 21.

AMDmi3 (changes requested):
Also multiple template syntax errors.

AMDmi3:
Timeout

repos.d/deb/devuan.yaml
hlein: Hah, I was overzealous when copying `debian.yaml`. Simplified/cleaned up.

repos.d/deb/devuan.yaml
hlein: Agreed; `debian.yaml` has it split out but that's only because they use a separate host. No sign of Devuan doing so, yet, so I squashed out this and other complexities.

PR #1589, opened 2026-05-06
Add stonking, remove plucky

arif-ali:
Update dates on release dates and expiration based on https://ubuntu.com/about/release-cycle

PR #1592, opened 2026-05-11
Add Ubuntu 26.10 and remove Ubuntu 25.04

jbicha:
per news section of https://packages.ubuntu.com/

and

https://salsa.debian.org/debian/distro-info-data/-/blob/main/ubuntu.csv


jbicha:
I noticed after I submitted this that there is a similar merge request already.

I consider distro-info-data to be the authoritative source for End of Life dates for Debian and Ubuntu. (For one thing, it gives a day instead of a month.) I have asked the Ubuntu Release Team to fix the discrepancy for the Legacy End of Life dates.

AMDmi3:
Duplicate of #1589

PR #1593, opened 2026-05-12
nix: track nixpkgs-unstable instead of nixos-unstable

mweinelt:
The nixpkgs-unstable branch better reflects package versions built than the nixos-unstable branch can. This is because we run lots of tests on nixos-unstable and it often has multiple days of delay.

Currently, nixpkgs-unstable gets evaluated every 8 hours or 3 times per day. The major difference here is that nixpkgs also contains packages built for Darwin.

https://hydra.nixos.org/jobset/nixpkgs/unstable (vs https://hydra.nixos.org/jobset/nixos/unstable)
https://channels.nixos.org/nixpkgs-unstable
https://github.com/NixOS/nixpkgs/tree/nixpkgs-unstable

This chart shows the age before a new channel versions appears for both:

image

On average this will update at least once a day down from every two to three days.

### DIFF
--- a/repos.d/nixos.yaml
+++ b/repos.d/nixos.yaml
@@ -62,7 +62,7 @@
     - name: packages-unstable.json
       fetcher:
         class: FileFetcher
-        url: https://channels.nixos.org/nixos-unstable/packages.json.br
+        url: https://channels.nixos.org/nixpkgs-unstable/packages.json.br
       parser:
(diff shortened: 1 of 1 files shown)


acid-bong:
maybe switch stable branches as well? just in case any Mac users are on stable

mweinelt:
There is no comparable branch for stable.

PR #1594, opened 2026-05-19
Add XRepo (xmake-repo) repository support

gistrec:
## Summary

Adds support for [xrepo / xmake-repo](https://github.com/xmake-io/xmake-repo), the C/C++ package repository for the [xmake](https://xmake.io/) build system. Similar in spirit to the vcpkg and conan integrations.

This is an alternative to #1585 — it consumes the pre-built static index that xmake-repo publishes specifically for Repology:

  - https://xmake-io.github.io/xmake-repo/repology_packages_index.json

The JSON already contains everything we need per package (`name`, `version`, `description`, `license`, `homepage`, `repository_url`, `download_url`), so we just reuse the existing `FileFetcher` instead of writing a custom fetcher that fans out to a separate API + per-package `xmake.lua` scrapes.

Verified end-to-end against the live index — **1905 packages** parse cleanly (with 1588 licenses, 1894 homepages, 1690 repository links, 1455 download links).

## Changes

- `repos.d/xrepo.yaml` — repository definition, points `FileFetcher` at the static JSON
- `repology/parsers/parsers/xrepo.py` — `XrepoJsonParser` streaming the `packages[]` array via `iter_json_list`
- `repology/packagemaker/names.py` — `XREPO_NAME` (maps to `GENERIC_SRC_NAME`)
- `repology-schemacheck.py` — adds `xrepo` to known families
- `testdata/xrepo.state` — small sample index for local checks

`xrepo_package_path` (used in `PACKAGE_SOURCES` / `PACKAGE_RECIPE` / `PACKAGE_RECIPE_RAW` link templates) is derived from the `packages/{first-letter}/{name}/` convention used by xmake-repo, since the static index doesn't include it. All three generated link types resolve against the live `dev` branch.

Leading `v` is stripped from versions like `v3.4.4` → `3.4.4`. `.git` `download_url`s (e.g. `…/repo.git`) are not added as `UPSTREAM_DOWNLOAD` links since they're VCS URLs, not archives.

### DIFF
--- a/repology/parsers/parsers/xrepo.py
+++ b/repology/parsers/parsers/xrepo.py
@@ -1,0 +1,58 @@
+# Copyright (C) 2026 Dmitry Marakasov <amdmi3@amdmi3.ru>
+#
+# This file is part of repology
+#
+# repology is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# repology is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with repology.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Iterable
+
+from repology.package import LinkType
+from repology.packagemaker import NameType, PackageFactory, PackageMaker
+from repology.parsers import Parser
+from repology.parsers.json import iter_json_list
+
+
+def _normalize_version(version: str) -> str:
+    if version.startswith('v') and len(version) > 1 and version[1].isdigit():
+        return version[1:]
+    return version
+
+
+class XrepoJsonParser(Parser):
+    def iter_parse(self, path: str, factory: PackageFactory) -> Iterable[PackageMaker]:
+        for pkgdata in iter_json_list(path, ('packages', None)):
+            name = pkgdata.get('name')
+            version = pkgdata.get('version')
+            if not name or not version:
+                continue
+
+            with factory.begin(name) as pkg:
+                pkg.add_name(name, NameType.XREPO_NAME)
+                pkg.set_version(version, _normalize_version)
+                pkg.set_extra_field('xrepo_package_path', f'packages/{name[0]}/{name}')
+
+                if (summary := pkgdata.get('description')):
+                    pkg.set_summary(summary)
+                if (homepage := pkgdata.get('homepage')):
+                    pkg.add_links(LinkType.UPSTREAM_HOMEPAGE, homepage)
+                if (license_ := pkgdata.get('license')):
+                    pkg.add_licenses(license_)
+                if (repo_url := pkgdata.get('repository_url')):
+                    pkg.add_links(LinkType.UPSTREAM_REPOSITORY, repo_url)
+
+                download_url = pkgdata.get('download_url')
+                if download_url and not download_url.endswith('.git'):
+                    pkg.add_links(LinkType.UPSTREAM_DOWNLOAD, download_url)
+
+                yield pkg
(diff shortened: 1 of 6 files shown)


AMDmi3:
Should we discuss data problems here, or in related xmake issue?

gistrec:
> Should we discuss data problems here, or in related xmake issue?

Let's continue in the related xmake-repo issue: https://github.com/xmake-io/xmake-repo/issues/9966

AMDmi3:
For the record, apart from the listed problems and a bunch of fake date based versions which can be ignored, the parser works and the data is mostly good.

2026-05-20-220222_sel

AMDmi3:
Timeout

gistrec:
Hi @AMDmi3! The blocker on the xmake-repo side is finally resolved, so I'd like to revive this — could you reopen the PR? (GitHub doesn't let me reopen it myself since it was closed by a maintainer.)

What changed since the timeout:
- xmake-io/xmake-repo#10355 was merged (2026-07-05): the index generator now properly resolves versions for alias, versionfile and host-conditional packages — the data problems discussed in xmake-io/xmake-repo#9966.
- The index at https://xmake-io.github.io/xmake-repo/repology_packages_index.json is live and regenerated hourly (currently 1971 packages).

I've updated the parser for the extended schema (pushed on top, no force-push):
- all available versions are emitted (`versions[]`), with download links (incl. mirrors) attached to the version they belong to;
- patch files are exposed as `PACKAGE_PATCH`/`PACKAGE_PATCH_RAW` links;
- alias entries (`base` field, e.g. `curl` → `libcurl`) are skipped;
- testdata updated to the current schema, and xrepo added to the `have_testdata` group so the fixture is actually exercised by `test_parsers_regress`.

Sanity check against the full live index: 1971 packages → 1946 projects (25 aliases skipped), 7948 version entries, no parse errors; flake8/mypy/schemacheck clean.

Happy to open a fresh PR instead if you'd prefer that.

repology/parsers/parsers/xrepo.py
```diff
@@ -0,0 +1,62 @@
...
+# along with repology.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Iterable
+
+from repology.package import LinkType
+from repology.packagemaker import NameType, PackageFactory, PackageMaker
+from repology.parsers import Parser
+from repology.parsers.json import iter_json_list
+
+_BLOB_URL_PREFIX = 'https://github.com/xmake-io/xmake-repo/blob/dev/'
```
AMDmi3: Hardcoded URLs are not acceptable in parsers. You need to move these into config and, if needed, pass missing fields via set_extra_field

repology/parsers/parsers/xrepo.py
```diff
@@ -0,0 +1,62 @@
...
+        for pkgdata in iter_json_list(path, ('packages', None)):
+            if 'base' in pkgdata:
+                # alias reusing another package's definition (e.g. curl → libcurl)
+                continue
+
+            name = pkgdata['name']
+
+            with factory.begin(name) as pkg:
+                pkg.add_name(name, NameType.XREPO_NAME)
+                pkg.set_extra_field('xrepo_package_path', f'packages/{name[0]}/{name}')
```
AMDmi3: You probably don't need this - in packagelinks, you can refer to the package name as `{srcname}` and to the package name first letter as `{srcname|first_letter}` (see aerynos.yaml as example)

repology/parsers/parsers/xrepo.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class XrepoJsonParser(Parser):
+    def iter_parse(self, path: str, factory: PackageFactory) -> Iterable[PackageMaker]:
+        for pkgdata in iter_json_list(path, ('packages', None)):
+            if 'base' in pkgdata:
+                # alias reusing another package's definition (e.g. curl → libcurl)
+                continue
+
+            name = pkgdata['name']
```
AMDmi3: Since this can technically throw, this needs to be inside the context manager as well. It picks the name added by `add_name`, so you don't need to pass it to factory.begin.

But since `xrepo_package_path` should also be removed, you don't need to introduce `name` at all, just use `add_name(pkgdata['name'])`

AMDmi3 (changes requested):
Also please squash into single commit.

gistrec:
Thanks for the review! All points addressed and squashed into a single commit:
- patch URLs moved out of the parser: paths are passed via `patch` extra field, URL templates now live in packagelinks config
- dropped `xrepo_package_path` extra field in favor of `packages/{srcname|first_letter}/{srcname}`
- `pkgdata['name']` moved inside the factory.begin() context manager

PR #1602, opened 2026-05-26
ditch some EOLed branch, and adopt to new source hosting platform

fundawang:
- openEuler has moved to new source hosting platform `atomgit`
- ditch EOLed branches
- add 24.03-LTS-SP3 which was released Dec 2025.

P.S., `389-ds-base` repo was renamed today, the reason why `three-eight-nine-ds-base` was used is that our old hosting platform (gitee) is rejecting project names beginning with number. More tweaks follow.

### DIFF
--- a/repos.d/rpm/openeuler.yaml
+++ b/repos.d/rpm/openeuler.yaml
@@ -27,24 +27,19 @@
       url: https://openeuler.org/en/
   packagelinks:
     - type: PACKAGE_SOURCES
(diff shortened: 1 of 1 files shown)


AMDmi3:
- We do not accept repository removals,
- No new versions are accepted unless the problem with inaccessible package sources is fixed.

We can only update the urls for now.

AMDmi3:
Timeout

Issue #1346, opened 2023-08-02
Salix is missing most repositories


gapan:
I can see there is already an entry for Salix, but it only includes a single repository for version 14.2 and only for the x86_64 architecture, therefore you are missing 10s of thousands of packages.

Issues:
* Version 15.0 has been released since almost a year now, if there should only be one version, it should be that. But ideally, at least versions 15.0 and 14.2 should be covered (and there are even older version 14.1 and 14.0 that are still supported).
* There is also a i486 architecture, which you are not listing at all.
* But most importantly, there are multiple repositories for each version and architecture. As it is, you are parsing the PACKAGES.json file in a single repository. For version 15.0 and architecture x86_64, the complete list of respective json files to parse is:
```
https://download.salixos.org/x86_64/15.0/PACKAGES.json
https://download.salixos.org/x86_64/extra-15.0/PACKAGES.json
https://download.salixos.org/x86_64/slackware-15.0/PACKAGES.json
https://download.salixos.org/x86_64/slackware-15.0/patches/PACKAGES.json
https://download.salixos.org/x86_64/xfce4.18-15.0/PACKAGES.json
```

Similarly for version 15.0 and architecture i486:
```
https://download.salixos.org/i486/15.0/PACKAGES.json
https://download.salixos.org/i486/extra-15.0/PACKAGES.json
https://download.salixos.org/i486/slackware-15.0/PACKAGES.json
https://download.salixos.org/i486/slackware-15.0/patches/PACKAGES.json
https://download.salixos.org/i486/xfce4.18-15.0/PACKAGES.json
```

For version 14.2, x86_64:
```
https://download.salixos.org/x86_64/14.2/PACKAGES.json
https://download.salixos.org/x86_64/extra-14.2/PACKAGES.json
https://download.salixos.org/x86_64/slackware-14.2/PACKAGES.json
https://download.salixos.org/x86_64/slackware-14.2/patches/PACKAGES.json
```

And for version 14.2, i486:
```
https://download.salixos.org/i486/14.2/PACKAGES.json
https://download.salixos.org/i486/extra-14.2/PACKAGES.json
https://download.salixos.org/i486/slackware-14.2/PACKAGES.json
https://download.salixos.org/i486/slackware-14.2/patches/PACKAGES.json
```

AMDmi3:
Upstream (homepage or download) links and links to package recipes (slackbuilds in this case, I think) are required for new submissions. As far as I can see salix does not provide these.

[...]

PR #1608, opened 2026-06-14
Expand Salix coverage: add 15.0, i486, extra/xfce subrepos

andy5995:
This should expand Salix coverage:

- Adds Salix 15.0 (previously only 14.2 was tracked).
- Adds the i486 architecture alongside x86_64 for both 14.2 and 15.0 (resolves the existing `# XXX: add 32 bit repo` note in `salix.yaml`).
- Adds the Salix-native `extra` sub-repository for both versions, and
  the `xfce` sub-repository (`xfce4.18`) for 15.0.

This is related to #1346 (cc @gapan) but a narrower scope. I didn't exercise the database write step locally.

### DIFF
--- a/repos.d/salix.yaml
+++ b/repos.d/salix.yaml
@@ -1,6 +1,35 @@
 ###########################################################################
 # Salix
 ###########################################################################
+
+{% set salix_mirror = "https://download.salixos.org" %}
+
+- name: salix_15_0
+  type: repository
+  desc: Salix 15.0
+  family: salix
+  ruleset: salix
+  color: '7ad37a'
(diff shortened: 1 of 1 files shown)


AMDmi3:
I don't see any issues raised in #1346 (also mentioned on [repository page](https://repology.org/repository/salix_14_2)) fixed.

andy5995:
> I don't see any issues raised in #1346 (also mentioned on [repository page](https://repology.org/repository/salix_14_2)) fixed.

Ah, yes, I see I missed some key info. Feel free to close if you don't think this PR will be needed as-is.

PR #1614, opened 2026-07-12
Fix reactos parser on Python 3.12+ (ConfigParser.readfp removed)

Teko012:
ConfigParser.readfp() was deprecated in Python 3.2 and removed in Python 3.12. On 3.12+ the reactos (rapps) parser raises AttributeError: 'ConfigParser' object has no attribute 'readfp', which fails parsing and drops the entire reactos repository.

Replace it with read_file() - the direct, semantically identical replacement available since Python 3.2. The # type: ignore is dropped since read_file() is properly typed.


AMDmi3:
Thanks! This is very timely because I was just about to update FreeBSD packages which were bumped to python 3.12

PR #1615, opened 2026-07-12
openvsx: paginate search API (?size=10000 now returns 400)

Teko012:
OpenVSX now rejects large ?size (max 1000), so the single-shot fetch 400s and the whole repo drops. Page by offset with a stable timestamp sort; write pages verbatim like the crates.io fetcher.

### DIFF
--- a/repology/fetchers/fetchers/openvsx.py
+++ b/repology/fetchers/fetchers/openvsx.py
@@ -1,0 +1,85 @@
+# Copyright (C) 2024 Dmitry Marakasov <amdmi3@amdmi3.ru>
+#
+# This file is part of repology
+#
+# repology is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


Teko012:
@AMDmi3 What do you think about this?

AMDmi3:
Sorry, I hope to get to reviewing it soon

AMDmi3:
The code is good, but the data looks to be truncated (it only returns 10k entries), and the results are sorted in a way the latest versions are lost. It looks like `query` (or `v2/query`) endpoint should be used instead (with `includeAllVersions=false`). It does not seem to be a way to get a consistent snapshot of the repository, as added/removed entries would break the pagination. We could try to implement a (not completely reliable) safeguard against it by ensuring that each page returns the same `totalSize`, and failing the fetch if it has not. It would be more or less acceptable if it works this way. Otherwise, we'll have to drop this source.

PR #1622, opened 2026-07-20
pypi: skip release versions containing newlines

Teko012:
## What

Skip PyPI release versions whose version string contains a `\n` or `\r`, logging a warning rather than producing a malformed package entry.

## Why

Some packages on PyPI expose release keys with embedded newline/carriage-return characters in the version string. These leak through the parser and cause malformed package data downstream. Filtering them at the parse stage keeps the rest of the package's releases intact while dropping only the bad entries.

### DIFF
--- a/repology/parsers/parsers/pypi.py
+++ b/repology/parsers/parsers/pypi.py
@@ -95,6 +95,10 @@ class PyPiCacheJsonParser(Parser):
                             pkg.add_links(link_type, url)
 
                 for version, releasedatas in pkgdata['releases'].items():
+                    if "\n" in version or "\r" in version:
+                        pkg.log(f"Skipping malformed PyPI release version {version!r}", severity=Logger.WARNING)
+                        continue
+
                     verpkg = pkg.clone()
 
                     verpkg.set_version(version)


AMDmi3:
Any examples?

Teko012:
Background: After https://github.com/repology/repology-rs/issues/537 was unsolvable, I decided to self-host this service, and create our own. I'm not sure why, but there is a discrepancy in the scraped data: it reports 455321 packages (yours) vs 827785 (self-hosted).

Processing our version of the dataset throws the error (with some added logging now to find which package): 
```
package='blargs'
version='0.2.4b\n0.2.5b'
RuntimeError: Attempt to spawn Package with unset version`
```
I guess the package/version in question: https://pypi.org/project/blargs/0.2.4b%0A0.2.5b/

Maybe the hosted dataset filters this out? I'm curious where the discrepancy comes from; maybe the git repo is not up to date with the latest version.

AMDmi3:
That's because pypicache had not seeded with all packages from pypi, only ones intersecting with repology. I don't see any need to fix this unless it fires on Repology. In which way it would be more correct to fix it in pypicache instead.

Teko012:
@AMDmi3 Is that a feature that I missed, or is that part not published?

AMDmi3:
If you're asking about not seeding pypicache, that would be `--no-bootstrap`.

Teko012:
Yeah, but how does it correlate with it with repology, per `only ones intersecting with repology.`? That's the part I wasn't sure.

AMDmi3:
Oneshot manual action with psql, like select python:* from repology | insert into pypicache queue

